PR #2, opened 2017-12-12
[WIP][11.0][add] server environment config settings

grindtildeath:


grindtildeath:
As such a module would be too tricky matching fields on res.config.settings with fields of their own models, this intended feature was replaced by https://github.com/OCA/server-env/pull/3 to match our needs.

PR #3, opened 2017-12-13
[11.0][Add] Module auth_oauth_environment

grindtildeath:


auth_oauth_environment/models/auth_oauth_provider.py
```diff
@@ -0,0 +1,70 @@
...
+                    and serv_config.get(provider_section_name, 'enabled'):
+                enabled_provider_ids.append(provider.id)
+        return [('id', 'in', enabled_provider_ids)]
+
+    @api.multi
+    def _compute_simple_name(self):
+        for provider in self:
+            provider.provider_simple_name = provider.name.split(' ')[0].lower()
+
+    @api.one
```
yvaucher: This will fail as when calling it you will get a list, prefer to use `@api.multi` and use a `self.ensure_one()`

auth_oauth_environment/models/auth_oauth_provider.py
```diff
@@ -0,0 +1,70 @@
...
+
+    managed_by_env = fields.Boolean(compute='_compute_server_env')
+    provider_simple_name = fields.Char(compute='_compute_simple_name')
+    client_id = fields.Char(string='Client ID', compute='_compute_server_env')
+    enabled = fields.Boolean(string='Allowed', compute='_compute_server_env',
+                             search='_search_enabled')
+
+    def _compute_server_env(self):
+        for provider in self:
+            provider_section_name = provider._get_provider_section_name()
```
yvaucher, line 23: Here you currently receive a list, see comment on method `_get_provider_section_name`

yvaucher (changes requested):

yvaucher:
please find a fix here https://github.com/grindtildeath/server-env/pull/1

yvaucher (approved):

gurneyalex (approved):
code review

auth_oauth_environment/models/auth_oauth_provider.py
gurneyalex: @yvaucher @grindtildeath  is this still relevant?

gurneyalex:
@grindtildeath could you rebase this one to get a fresh clean build ?

grindtildeath:
There's a weird travis error here as it seems to run on Odoo 13.0 although this PR targets branch 11.0:
 - https://travis-ci.org/OCA/server-env/jobs/596472536#L231
 - https://github.com/grindtildeath/server-env/blob/11_0_add_auth_oauth_environment/.travis.yml#L17

auth_oauth_environment/models/auth_oauth_provider.py
```diff
@@ -0,0 +1,71 @@
...
+import json
+from lxml import etree
+
+from odoo import fields, models, api, _
+from odoo.addons.server_environment import serv_config
+
+
+class AuthOAuthProvider(models.Model):
+
+    _inherit = 'auth.oauth.provider'
```
simahawk, line 13: wouldn't be better to use the new mixin?
https://github.com/OCA/server-env/tree/11.0/server_environment#usage

grindtildeath: The mixin was introduced after this PR was done, so IMO we could merge it as is and it could be improved later on.

PR #5, opened 2017-12-18
Server environment ir config parameter

TDu:


simahawk:
@TDu the log seems polluted w/ `auditolog` commits.
Also, you can now remove the pending dependency on server_environment

TDu:
Replaced by #8

PR #8, opened 2018-01-16
[MIG][11.0] server_environment_ir_config_parameter

simahawk:
replaces https://github.com/OCA/server-env/pull/5

NOTE: I used `git filter-branch  --prune-empty --subdirectory-filter ...` to move the module from server-tools which ended up in the root of the repo. I moved it to the right place in the migration commit.
I don't think it's a problem tho.


simahawk:
@sbidoul ping :)

TDu (approved):
Thank you for taking over. LGTM

pedrobaeza:
You can use `git am` the same as in migration guide. The only difference is the remotes you configure and put in the comment.

simahawk:
@pedrobaeza funny thing is that I tried that way and it didn't worked... but I guess I messed up w/ remotes :/
I'll try it again.

pedrobaeza:
Well, as you already have it, no need to try more, but I pointed out only as reference.

sbidoul (approved):
(code review)

.travis.yml
```diff
@@ -10,6 +10,7 @@ addons:
     packages:
       - expect-dev  # provides unbuffer utility
       - python-lxml # because pip installation is slow
+      - unixodbc-dev
```
leemannd: Why is it required?

leemannd (commented):
LGTM
I only have a question.

.travis.yml
simahawk: Thierry added it to fix tests on py3. No traceback tho. I removed it, let's see how travis behaves...

simahawk: @leemannd tests pass anyway. Tnx :+1:

server_environment_ir_config_parameter/tests/config_param_test.xml
```diff
@@ -0,0 +1,8 @@
+<odoo>
+  <data>
```
leemannd: Isn't this useless if there is no noupdate?

leemannd (approved):
Non blocking comment.
LGTM! Code review only

server_environment_ir_config_parameter/models/ir_config_parameter.py
```diff
@@ -0,0 +1,54 @@
...
+SECTION = 'ir.config_parameter'
+
+
+class IrConfigParameter(models.Model):
+
+    _inherit = 'ir.config_parameter'
+
+    @api.model
+    def get_param(self, key, default=False):
+        value = super(IrConfigParameter, self).get_param(key, default=None)
```
mpanarin: supers can be reduced to `super().foo()` in python3 )

server_environment_ir_config_parameter/tests/test_server_environment_ircp.py
```diff
@@ -0,0 +1,80 @@
...
+
+from odoo.exceptions import UserError
+from odoo.tests import common
+from odoo.tools import convert_file
+from odoo.modules.module import get_resource_path
+
+
+class TestEnv(common.TransactionCase):
+
+    def setUp(self):
```
mpanarin, line 12: Isn't setupclass just better?

mpanarin (approved):
other than small things
LGTM

server_environment_ir_config_parameter/tests/test_server_environment_ircp.py
simahawk: no, if you don't do anything :)
setUpClass must be used if you are creating several records that are used by each test.

PR #9, opened 2018-02-21
[FIX] server_environment_file does not exist as a packaged addon

benwillig:


sbidoul (approved):
A straightforward port of previous versions

JordiBForgeFlow (approved):

PR #12, opened 2018-07-19
Add SERVER_ENV_CONFIG to configure vars from env. variable

guewen:
It allows to inject secret values without storing them in files, change values outside of the "project distribution" (docker, git, ...).

### DIFF
--- a/server_environment/README.rst
+++ b/server_environment/README.rst
@@ -47,8 +47,12 @@ You should then edit the settings you need in the
   `default/` directory using the .ini file syntax;
 * each environment you need to define is stored in its own directory
   and can override or extend default values;
-* finally, you can override or extend values in the main configuration
-  file of you instance.
+* you can override or extend values in the main configuration
+  file of you instance;
+* additional configuration can be passed in the environment variable
+  ``SERVER_ENV_CONFIG``, overriding any values set in the configuration files.
+  This is a multi-line environment variable in the same configparser format than
+  the files.
 
 Values associated to keys
 containing 'passw' are only displayed in the 'dev' environment.
--- a/server_environment/README.rst
+++ b/server_environment/README.rst
@@ -27,6 +27,8 @@ the incoming and outgoing mail servers depending on the environment.
 To install this module, you need to provide a companion module called
 `server_environment_files`. You can copy and customize the provided
 `server_environment_files_sample` module for this purpose.
+You can provide additional options in environment variables
+``SERVER_ENV_CONFIG`` and ``SERVER_ENV_CONFIG_SECRET``.
 
 
 Configuration
@@ -48,11 +50,11 @@ You should then edit the settings you need in the
 * each environment you need to define is stored in its own directory
   and can override or extend default values;
 * you can override or extend values in the main configuration
-  file of you instance;
+  file of your instance;
 * additional configuration can be passed in the environment variable
-  ``SERVER_ENV_CONFIG``, overriding any values set in the configuration files.
-  This is a multi-line environment variable in the same configparser format than
-  the files.
+  ``SERVER_ENV_CONFIG`` and/or ``SERVER_ENV_CONFIG_SECRET``, overriding any
+  values set in the configuration files. This is a multi-line environment
+  variable in the same configparser format than the files.
 
 Values associated to keys
 containing 'passw' are only displayed in the 'dev' environment.
--- a/server_environment/README.rst
+++ b/server_environment/README.rst
@@ -35,29 +35,76 @@ Configuration
 =============
 
 To configure this module, you need to edit the main configuration file
-of your instance, and add a directive called `running_env`. Commonly
+of your instance, and add a directive called ``running_env``. Commonly
 used values are 'dev', 'test', 'production'::
 
   [options]
   running_env=dev
 
-You should then edit the settings you need in the
-`server_environment_files` addon. The
-`server_environment_files_sample` can be used as an example:
+Values associated to keys containing 'passw' are only displayed in the 'dev'
+environment.
+
+You have several possibilities to set configuration values:
+
+server_environment_files
+------------------------
+
+You can edit the settings you need in the ``server_environment_files`` addon. The
+``server_environment_files_sample`` can be used as an example:
 
 * values common to all / most environments can be stored in the
-  `default/` directory using the .ini file syntax;
+  ``default/`` directory using the .ini file syntax;
 * each environment you need to define is stored in its own directory
   and can override or extend default values;
 * you can override or extend values in the main configuration
   file of your instance;
-* additional configuration can be passed in the environment variable
-  ``SERVER_ENV_CONFIG`` and/or ``SERVER_ENV_CONFIG_SECRET``, overriding any
-  values set in the configuration files. This is a multi-line environment
-  variable in the same configparser format than the files.
 
-Values associated to keys
-containing 'passw' are only displayed in the 'dev' environment.
+Environment variable
+--------------------
+
+You can define configuration in the environment variable ``SERVER_ENV_CONFIG``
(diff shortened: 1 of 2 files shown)


server_environment/README.rst
```diff
@@ -47,8 +47,12 @@ You should then edit the settings you need in the
   `default/` directory using the .ini file syntax;
 * each environment you need to define is stored in its own directory
   and can override or extend default values;
-* finally, you can override or extend values in the main configuration
+* you can override or extend values in the main configuration
   file of you instance.
+* additional configuration can be passed in the environment variable
+  ``SERVER_ENV_CONFIG``, overriding any values set in the configuration files.
+  This is a multi-line environment variable in the same configparser format than
+  the files
```
elicoidal: add period

elicoidal (approved):
LGTM

vrenaville (approved):

server_environment/README.rst
guewen: added

server_environment/serv_config.py
```diff
@@ -97,9 +97,20 @@ def _load_config():
...
     config_p.remove_section('options')
 
+    env_config = os.getenv('SERVER_ENV_CONFIG')
+    if env_config:
+        try:
+            config_p.read_string(env_config)
+        except configparser.Error as err:
+            raise Exception(
+                'SERVER_ENV_CONFIG content could not be parsed: %s'
+                % (err,)
```
sbidoul: Have you checked `err` does not reveal the content of the of the environment variable?

sbidoul:
Excellent!

server_environment/serv_config.py
guewen: Good catch, it does!

```
2018-07-19 06:18:57,343 1 CRITICAL odoodb odoo.modules.module: Couldn't load module server_environment
2018-07-19 06:18:57,343 1 CRITICAL odoodb odoo.modules.module: SERVER_ENV_CONFIG content could not be parsed: File contains no section headers.
file: '<string>', line: 1
'my super secret but invalid format'
2018-07-19 06:18:57,345 1 WARNING odoodb odoo.modules.loading: Transient module states were reset
2018-07-19 06:18:57,345 1 ERROR odoodb odoo.modules.registry: Failed to load registry
Traceback (most recent call last):
  File "/opt/odoo/external-src/server-env/server_environment/serv_config.py", line 107, in _load_config
    config_p.read_string(env_config)
  File "/usr/lib/python3.5/configparser.py", line 720, in read_string
    self.read_file(sfile, source)
  File "/usr/lib/python3.5/configparser.py", line 715, in read_file
    self._read(f, source)
  File "/usr/lib/python3.5/configparser.py", line 1077, in _read
    raise MissingSectionHeaderError(fpname, lineno, line)
configparser.MissingSectionHeaderError: File contains no section headers.
file: '<string>', line: 1
'my super secret but invalid format'
```

I'll see how to prevent that.

max3903 (approved):

guewen:
@sbidoul I tested a range of potential errors and the config printed is minimal, the most sensitive one is the first below, if you miss a section header, it shows the first line. We could show only the class name in the message (`<class configparser.DuplicateSectionError>`, `<class configparser.MissingSectionHeaderError>`, ...), should I do that?

```
Exception: SERVER_ENV_CONFIG content could not be parsed: File contains no section headers.
file: '<string>', line: 1
'sftp_password=password\n'
```


```
Exception: SERVER_ENV_CONFIG content could not be parsed: Source contains parsing errors: '<string>'
        [line  2]: 'sftp_port\n'
```

```
Exception: SERVER_ENV_CONFIG content could not be parsed: While reading from '<string>' [line  3]: option 'sftp_port' in section 'storage_backend.SFTP Backend' already exists
```

```
Exception: SERVER_ENV_CONFIG content could not be parsed: While reading from '<string>' [line  6]: section 'ir_mail_server' already exists
```

```
configparser.InterpolationSyntaxError: bad interpolation variable reference '%(test)'
```

```
configparser.InterpolationMissingOptionError: Bad value substitution: option 'type' in section 'incoming_mail.foo' contains an interpolation key 'test' which is not a valid option name. Raw value: 'imap%(test)s'
```

sbidoul:
It's good. I think I raised that more out of curiosity than anything else, and I would not have thought about it if you had not written that try/except. If something had to be done that would be in ConfigParser I guess.

sbidoul (approved):

server_environment/README.rst
```diff
@@ -47,8 +47,12 @@ You should then edit the settings you need in the
   `default/` directory using the .ini file syntax;
 * each environment you need to define is stored in its own directory
   and can override or extend default values;
-* finally, you can override or extend values in the main configuration
-  file of you instance.
+* you can override or extend values in the main configuration
+  file of you instance;
```
TDu: missing 'r' : your instance ?

TDu (approved):
:+1: 
Just a small typo maybe.

server_environment/README.rst
guewen: fixed

guewen:
@sbidoul @max3903 @TDu @elicoidal @vrenaville 

Thanks for your reviews. While setting up the deploy using this, I observed that we split our configuration options in 2 parts: most of the configuration for a service is in plain-text - because it's convenient to see them in clear - and the rest of the configuration is secrets which are encrypted. For this reason, I added a `SERVER_ENV_CONFIG_SECRET`, which is handled the same way than `SERVER_ENV_CONFIG`, it's only a convenience for the deploy.
 I hope you won't mind my additional commit.

server_environment/serv_config.py
```diff
@@ -97,9 +99,21 @@ def _load_config():
         config_p.read(conf_files)
     except Exception as e:
         raise Exception('Cannot read config files "%s":  %s' % (conf_files, e))
+
     config_p.read(system_base_config.rcfile)
     config_p.remove_section('options')
 
+    for varname in ENV_VAR_NAMES:
```
elicoidal, line 106: @guewen When reading the code, I am not sure what is the value of splitting.
You do not log any of them anymore so no risk of showing secret content.
I would understand that the main advantage is to make public or default one while the other is set up in a specific file in a server.
If that is your view, I would suggest to add an explanation in the README about how to use both variables

guewen: > I would understand that the main advantage is to make public or default one while the other is set up in a specific file in a server.

Yes, specifically we have something like this in our deployments:

``rancher.public.env`` is a file which contains all the public environment variables in clear text:

```
export WORKERS='8'
export MAX_CRON_THREADS='1'
export LOG_LEVEL=info
export LOG_HANDLER=":INFO"
export DB_MAXCONN=5

# server environment
export SERVER_ENV_CONFIG="
[storage_backend.my-sftp]
sftp_server=10.10.10.10
sftp_login=foo
sftp_port=22200
directory_path=Odoo
"
```

``rancher.env.gpg`` which is encrypted and contains secrets:

```
export DB_PASSWORD='xxxxxxxxx'
export KEYCHAIN_KEY='egS_Oxxxxxxxxxxxxxxxxxxxxxxxxxxx='
export SERVER_ENV_CONFIG_SECRET="
[storage_backend.my-sftp]
sftp_password=xxxxxxxxx
"
```

I think this is a common pattern in deployment, furthermore with Docker now. If we wanted to support only the `SERVER_ENV_CONFIG`, we would have to merge the content of the the 2 files beforehand which would be a pain...

I'll add a few lines more in the documentation.

elicoidal: :+1: We have similar pattern too.

guewen: @elicoidal does my last commit improves the understanding?

elicoidal: Perfect!

PR #13, opened 2018-07-19
server env: add fallback on database default values

guewen:
# Server environment mixin

## Introduction
Today, every module using server_environment reimplements the same computed
functions and fields overrides. (examples: [mail_environment](https://github.com/OCA/server-env/blob/5cdc84821add7fefb85f2a117ca015d230f73ba8/mail_environment/models/fetchmail_server.py), [server_env_connector_magento](https://github.com/OCA/connector-magento/blob/72fd28f321f144d806207317e2122918b15da9c5/server_env_connector_magento/models/magento_backend.py))

This is a proposal of an abstraction which will reduce duplication, and bring new features along the way.

## Features

* Provide a mixin to automatically convert fields to env-computed fields.
* Provide default values stored in database when a field is not set in
configuration files
* Allow edition of env-computed fields not set in configuration files, they are stored
in a serialized field
* Fields set in configuration files remain read-only
* The default value of a field (`fields.Char(default='foo')`) is properly used when there is no key in configuration files
* Integration with keychain, instead of being stored in the serializable datastore, the default value can be configured to be stored in a keychain account (https://github.com/guewen/storage/blob/eec246529f05460f8077c0b08617af56e5525c53/server_env_storage_backend_sftp/models/storage_backend.py#L17-L23)

Ultimately, my goal is to make the `server_environment_files` optional:

* default / fallback values that can be shared across environment stored in database
* env-dependent configuration in SERVER_ENV_CONFIG/SERVER_ENV_CONFIG_SECRET environment variables (#12)
* as the fields which are not set in configuration files are editable, the integrator can decide which fields are dependent of the environment (set a key in the files) or shared (no key in the files, can edit in database and set a default value).

## Showcase

peek 2018-07-19 15-31

### Code before (example from mail_environment)

https://github.com/OCA/server-env/blob/5cdc84821add7fefb85f2a117ca015d230f73ba8/mail_environment/models/ir_mail_server.py#L8-L44

### Code after (example from mail_environment)

```python
class IrMailServer(models.Model):
    _name = "ir.mail_server"
    _inherit = ["ir.mail_server", "server.env.mixin"]

    @property
    def _server_env_fields(self):
        base_fields = super()._server_env_fields
        mail_fields = {
            "smtp_host": {},
            "smtp_port": {
                "getter": "getint",
            },
            "smtp_user": {},
            "smtp_pass": {},
            "smtp_encryption": {},
        }
        mail_fields.update(base_fields)
        return mail_fields

    @api.model
    def _server_env_global_section_name(self):
        """Name of the global section in the configuration files

        Can be customized in your model
        """
        return 'outgoing_mail'
```


guewen:
> Ultimately, my goal is to make the server_environment_files unnecessary:

well, with that we don't need anymore `running_env` so we can wonder if it should be a new module, but by convenience it can still be fine to store a few options in server_environment_files (ribbon.name I name you)

server_environment/__init__.py
```diff
@@ -17,4 +17,9 @@
 #    along with this program.  If not, see <http://www.gnu.org/licenses/>.
 #
 ##############################################################################
+from . import models
+# Add an alias to access to the 'serv_config' module as it is shadowed
+# the next line by an import of a variable with the same name.
+# We can't change the import of serv_config for backward compatibility.
```
yvaucher: Add a comment to fix it on Odoo 12 migration?

server_environment/__manifest__.py
```diff
@@ -20,8 +20,11 @@
 
 {
     "name": "server configuration environment files",
-    "version": "11.0.1.0.1",
-    "depends": ["base"],
+    "version": "11.0.1.2.0",
```
yvaucher: I would put it as a major version. While it doesn't break compatibility, the paradigm is reversed as it sort of deprecate `server_environment_files`

server_environment/models/server_env_mixin.py
```diff
@@ -0,0 +1,384 @@
...
+    By default, it looks for the configuration in a section named
+    ``[model_name.Record Name]`` where ``model_name`` is the ``_name`` of the
+    model with ``.`` replaced by ``_``. Then in a global section which is only
+    the name of the model. They can be customized by overriding the method
+    :meth:`~_server_env_section_name` and
+    :meth:`~_server_env_global_section_name`.
+
+    For each field transformed to an env-computed field, a companion field
+    ``<field>_env_default`` is automatically created. When it's value is set
+    and the configuration files do not contain a key, the env-computed field
```
yvaucher: `a key` seems a bit too generic.

You might want to state `a matching key` or `a key for that field`

server_environment/models/server_env_mixin.py
```diff
@@ -0,0 +1,384 @@
...
+
+    def _server_env_default_fieldname(self, base_field_name):
+        """Return the name of the field with default value"""
+        options = self._server_env_fields[base_field_name]
+        if options.get('no_default_field'):
+            return ''
+        return '%s_env_default' % (base_field_name,)
+
+    def _server_env_is_editable_fieldname(self, base_field_name):
+        """Return the name of the field for "is editable
```
yvaucher: Missing a closing quote

server_environment/tests/testfiles/default/base.conf
```diff
@@ -0,0 +1,4 @@
+[external_service.ftp]
+host = sftp.example.com
+user = foo
+password = bar
```
yvaucher: Missing EOL

test_server_environment/__manifest__.py
```diff
@@ -0,0 +1,19 @@
...
+{
+    "name": "Automated tests for server environment - technical",
+    "summary": "Used to run automated tests, do not install",
+    "version": "11.0.1.0.0",
+    "depends": [
+        "server_environment",
+    ],
+    "author": "Camptocamp,Odoo Community Association (OCA)",
+    "website": "http://odoo-community.org/",
+    "license": "GPL-3 or any later version",
```
yvaucher: AGPL instead of GPL ?

yvaucher (approved):
Great!

Just some nitpickings which are non-blocking.

mail_environment/models/fetchmail_server.py
```diff
@@ -3,58 +3,46 @@
...
+        mail_fields = {
+            "server": {},
+            "port": {
+                "getter": "getint",
+            },
+            "type": {},
+            "user": {},
+            "password": {},
+            "is_ssl": {
+                "getter": "getbool",
```
sbidoul: could this be inferred from the field type?

guewen: of course!

server_environment/__init__.py
guewen: done

server_environment/__manifest__.py
guewen: done

server_environment/models/server_env_mixin.py
guewen: done

server_environment/models/server_env_mixin.py
guewen: done

server_environment/tests/testfiles/default/base.conf
guewen: done

mail_environment/models/fetchmail_server.py
guewen: done in 303e5e6

max3903 (approved):

mail_environment/models/fetchmail_server.py
sbidoul: cool, thanks!

sbidoul:
Great stuff!

Just a comment: please don't deprecate `running_env` and `server_environment_files`: it's super useful to provide sane defaults per environmenet when you need to send a release to a sysadmin that will deploy, possibly without reading the release notes. What could be done is requiring `running_env` only if `odoo.addons.server_environment_file` is importable.

guewen:
> Just a comment: please don't deprecate running_env and server_environment_files

:+1: I prefer to make them optional. I think we'll probably keep to use them at least for the "ribbon.name" ir config parameter per environment (web_ribbon)

guewen:
> What could be done is requiring running_env only if odoo.addons.server_environment_file is importable.

At the moment `running_env` is still required, even if `server_environment_files` is absent. On one hand, it could make sense to require it only if `server_environment_files` is there, but on the other hand, the `running_env` information could be a valuable information used in some kind of logic of an addon - we use it in our projects to check if the appropriate variables have been set in production for instance - the `running_env` could be required by the addon relying on it though. No strong opinion on that then! I might propose a subsequent pull request after this one for that.

server_environment/README.rst
```diff
@@ -106,17 +108,43 @@ A second file which is encrypted and contains secrets::
...
+--------------
+
+When using the ``server.env.mixin`` mixin, for each env-computed field, a
+companion field ``<field>_env_default`` is created. This field is not
+environment-dependent. It's a fallback value used when no key is set in
+configuration files / environment variable.
+
+When the default field is used, the field is made editable on Odoo.
+
+Note: a present key with an empty value do not fallback on the default field.
```
simahawk: "Note: empty env keys will take precedence over default fields anyway"?

mail_environment/models/ir_mail_server.py
```diff
@@ -1,44 +1,32 @@
...
+    _name = "ir.mail_server"
+    _inherit = ["ir.mail_server", "server.env.mixin"]
+
+    @property
+    def _server_env_fields(self):
+        base_fields = super()._server_env_fields
+        mail_fields = {
+            "smtp_host": {},
+            "smtp_port": {
+                "getter": "getint",
```
simahawk: isn't this automatic?

server_environment/models/server_env_mixin.py
```diff
@@ -0,0 +1,405 @@
...
+
+            @property
+            def _server_env_fields(self):
+                base_fields = super()._server_env_fields
+                sftp_fields = {
+                    "sftp_server": {
+                        "getter": "get",
+                    },
+                    "sftp_port": {
+                        "getter": "getint",
```
simahawk: I'd remove getters where not need and ad the optional example in case you have to convert to different values

server_environment/models/server_env_mixin.py
```diff
@@ -0,0 +1,405 @@
...
+        return value
+
+    @api.multi
+    def _server_env_has_key_defined(self, field_name):
+        self.ensure_one()
+        global_section_name = self._server_env_global_section_name()
+        section_name = self._server_env_section_name()
+        has_global_config = (
+            global_section_name in serv_config
+            and field_name in serv_config[global_section_name]
```
simahawk, line 205: `field_name in serv_config.get(global_section_name, {})`?

server_environment/models/server_env_mixin.py
```diff
@@ -0,0 +1,405 @@
...
+        )
+        has_config = (
```
simahawk, line 207: same?

server_environment/models/server_env_mixin.py
```diff
@@ -0,0 +1,405 @@
...
+
+    def _server_env_transform_field_to_read_from_env(self, field):
+        """Transform the original field in a computed field"""
+        field.compute = '_compute_server_env'
+
+        inverse_method_name = '_inverse_server_env_%s' % field.name
+        inverse_method = partialmethod(
+            ServerEnvMixin._inverse_server_env, field.name
+        )
+        setattr(ServerEnvMixin, inverse_method_name, inverse_method)
```
simahawk: shouldn't this be set on the model class instead?

server_environment/models/server_env_mixin.py
```diff
@@ -0,0 +1,405 @@
...
+            )
+            self._add_field(fieldname, field)
+
+    def _server_env_add_default_field(self, base_field):
+        """Add a field storing the default value
+
+        The default value is used when there is no key for an env-computed
+        field in the configuration files.
+
+        The field is stored in the serialized field ``server_env_defaults``.
```
simahawk: isn't it stored in db?

TDu (approved):

server_environment/models/server_env_mixin.py
```diff
@@ -0,0 +1,405 @@
...
+
+            if hasattr(base_field, 'selection'):
+                field_args['selection'] = base_field.selection
+            field = base_field_cls(**field_args)
+            self._add_field(fieldname, field)
+
+    @api.model
+    def _setup_base(self):
+        super()._setup_base()
+        for fieldname, options in self._server_env_fields.items():
```
simahawk: `for fieldname in self._server_env_fields.keys()`:

test_server_environment/README.rst
```diff
@@ -0,0 +1,46 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
```
simahawk, line 1: do we really need a separata package for testing?

simahawk (changes requested):

test_server_environment/README.rst
guewen: To have models, I don't like much creating tests models conditionally in the base addon (test addons is what's used in the core so I'm sure this is supported)

server_environment/models/server_env_mixin.py
guewen: How? `self.__class__` does not work. You might have a point though

server_environment/models/server_env_mixin.py
guewen: yes, stored in db in a serialized field, is it contradictory?

server_environment/models/server_env_mixin.py
sbidoul: @guewen you might indeed want to add a couple of tests with `_inherit` (different model name) and `_inherits` to check it works properly.

server_environment/models/server_env_mixin.py
simahawk: hmm isn't it stored in its own automatic field?
https://github.com/OCA/server-env/pull/13/files#diff-3e560ae060f8d61b271b31acc06ef3b1R271

server_environment/models/server_env_mixin.py
simahawk: `type(self)` does not work?

server_environment/models/server_env_mixin.py
guewen: @simahawk  the automatic field is a sparse field, which mean everything ends in ``server_env_defaults`` as json

server_environment/models/server_env_mixin.py
guewen: > @guewen you might indeed want to add a couple of tests with _inherit (different model name) and _inherits to check it works properly.

yes, with identical field names in both models, I'll add that

server_environment/models/server_env_mixin.py
guewen: ah no I already tried that, but it's not a dict, you get a `configparser.NoSectionError` exception when using `get` and the section does not exist

server_environment/models/server_env_mixin.py
guewen: same

server_environment/models/server_env_mixin.py
sbidoul: > type(self) does not work?

FWIW, the model class (as written by the developer) is `type(self).__bases__[0]`. In scobidoo I ended up injecting my fields in that class, in `_prepare_setup()` so they get properly propagated to children (included delegation children).

guewen: Thanks for the hint!

I added tests in 04e54e0, but the thing is that I could not have failing tests with `ServerEnvMixin` instead of `type(self)`. It's probably because that anyway the inverse method is doing the same thing so whether it's defined on ServerEnvMixin or the final model, as long as it's reachable, the result is the same. Anyway I changed it to be more explicit.

Regarding the delegation, I think I don't have the same issues than you, because:
* if you don't override the fields `_inherits` env-computed fields, they are handled by the "parent" model
* if you override some fields in the `_inherits` model, in that case you have to use the mixin again which will take care of the setup of these fields. I had to change these lines in the setup to "force" the override of the fields "env_default" and "env_is_editable" fields though

```
        if fieldname not in self._fields:
```
to
```
        if fieldname not in self._fields or self._fields[fieldname].inherited:
```

guewen:
The red build is not travis but runbot, I can't see why

Edit:

probably

```
2018-07-25 15:37:00,105 165 WARNING openerp_test odoo.models: Field server.env.test.inherits1.base_id with delegate=True must be required.
2018-07-25 15:37:00,130 165 WARNING openerp_test odoo.models: Field server.env.test.inherits2.base_id with delegate=True must be required.
```

guewen:
@sbidoul @simahawk could you recheck please? :hugs:

server_environment/models/server_env_mixin.py
simahawk: :open_mouth: :+1:

simahawk (approved):
LG

server_environment/tests/common.py
```diff
@@ -0,0 +1,59 @@
...
+
+from odoo.tests import common
+from odoo.addons.server_environment import server_env
+from odoo.tools.config import config
+
+import odoo.addons.server_environment.models.server_env_mixin as \
+    server_env_mixin
+
+
+class ServerEnvironmentCase(common.TransactionCase):
```
simahawk: SavepointCase not an option? I see records created for each tests. 
Not a big deal but we some time in our already big builds :innocent:

guewen: Done!

simahawk:
@max3903 squashing was needed :cry:

PR #20, opened 2019-01-04
[12.0][MIG] mail_environment from 11.0

sebalix:


sbidoul (approved):
(code review)

sbidoul:
@gurneyalex interesting behaviour of the CLA bot here.

mail_environment/__manifest__.py
```diff
@@ -0,0 +1,17 @@
...
+    'category': 'Tools',
+    'summary': 'Configure mail servers with server_environment_files',
+    'author': "Camptocamp, Odoo Community Association (OCA)",
+    'license': 'AGPL-3',
+    'website': 'https://github.com/OCA/server-env',
+    'depends': [
+        'fetchmail',
+        'server_environment',
+    ],
+    'data': [],
```
simahawk: this can be removed

mail_environment/readme/CREDITS.rst
```diff
@@ -0,0 +1 @@
+* Odoo Community Association: `Icon <https://github.com/OCA/maintainer-tools/blob/master/template/module/static/description/icon.svg>`_.
```
simahawk: I don't think this is needed

simahawk (approved):
LG, just some minor fixup

sbidoul:
@yvaucher @guewen could you review and merge this?

rvalyi:
@sbidoul @yvaucher @guewen Sorry for the off-topic but I have a question for server_env: is there some sort of support to target e specific database? Cause imagine if you are in a multi-tenant SaaS environment (the only way to do cheap SaaS, as Odoo SA does), can you set the different databases differently? What is the idea?

sbidoul:
@rvalyi multi-tenant is not going to work with server-env. The [good stuff happens at python import time](https://github.com/OCA/server-env/blob/11.0/server_environment/serv_config.py#L141), so it does not know anything about databases...

rvalyi:
@sbidoul @yvaucher @guewen that is what I thought... Then probably a module like keychain will still make sense to act as an encrypted secret vault for user data in multi-tenant context... Then I think it important to keep in mind that simple cases may be handled by server_env while it is interesting to keep multi-tenant compatibility in mind for more complex cases (are common abstractions required?). Finally may be you are not aware of this but it looks the guys from Modoolar (you know the nice project customization) made a module like keychain (they probably ignored keychain) https://github.com/modoolar/vault/tree/11.0/encrypted_vault
cc @hparfr @sebastienbeau @PierrickBrun

Overall a 1st approach is to tell Odoo is not responsible to keep secrets and we only pass them with ENV vars, or server_env, that's kind of the https://12factor.net/ clean scalable design, I like it.
But if Odoo doesn't keep hold secrets, then something else should... If you are not in a multi-tenant context, a simple file can do the job. But in a multi-tenant system you need a secret vault system...

If you think about a vault system you ca have 2 designs:

1) use Odoo itself as the vault system, that is the approach of keychain and Modoolar encrypted_vault I guess. Then we should ensure it work smoothly like server_env would...

2) Eventually you use an external dedicated vault system, as there are dozens of them. For instance Odoo SA could use their external account oauth provider as the vault system but there are existing tools too.

In such an external vault system, how do you forward the secret to the target Odoo database and modules?
Eventually, server_env could be designed in a way that it could act as THE mediator. Then whether you use keychain, Odoo SA account database or whatever, a single module like this one could set up your mail gateway for instance.

Just thinking loud for the future (not for this specific PR)... Was that clear? Any though?

gurneyalex (approved):

pedrobaeza:
@gurneyalex please remember to mark the check on the migration issue when merging migration PRs.

PR #22, opened 2019-05-13
ADD module carrier_environment

bealdav:


angelmoya (approved):
:+1:

angelmoya:
@jctrey check this pr to use as dependency in tracking link modules

angelmoya:
Maybe we need also selection field, like we use carrier_type.

Then we can set carrier_account on carrier_delivery, this way we can now carrier_type on pickings.

florian-dacosta (approved):

gurneyalex (approved):

PR #23, opened 2019-05-14
[12.0] Add new module data_encryption

florian-dacosta:
Simple module to encrypt and decrypt data from a key defined in the configuration file.
It is a low level module which do nothing by itself

A example of implementation has been made with server_environment_data_encryption module.
https://github.com/OCA/server-env/pull/24

@sebastienbeau @hparfr


data_encryption/models/encrypted_data.py
```diff
@@ -0,0 +1,130 @@
...
+            self = self.with_context(bin_size=False)
+        if self.env.user.id != SUPERUSER_ID:
+            raise AccessError(
+                _("Encrypted data can only be read as superuser"))
+        if not env:
+            env = self._retrieve_env()
+        encrypted_rec = self.search(
+            [('name', '=', name), ('environment', '=', env)])
+        if not encrypted_rec:
+            return ''
```
hparfr: return None

data_encryption/models/encrypted_data.py
```diff
@@ -0,0 +1,130 @@
...
+        required=True, readonly=True, index=True,
+        help="Technical name")
+    environment = fields.Char(
+        required=True,
+        index=True,
+        help="Concerned Odoo environment (prod, preprod...)")
+    encrypted_data = fields.Binary()
+
+    _sql_constraints = [
+        ('name_environment_uniq', 'unique (name, environment)',
```
hparfr, line 35: add company_id

data_encryption/models/encrypted_data.py
```diff
@@ -0,0 +1,130 @@
...
+        except (ValueError, TypeError):
+            raise ValidationError(
+                _("The data you are trying to read are not in a json format"))
+
+    @staticmethod
+    def _retrieve_env():
+        """Return the current environment
+           Raise if none is found
+        """
+        current = config.get('running_env') or False
```
hparfr: config.get('running_env', False)

data_encryption/__manifest__.py
```diff
@@ -0,0 +1,23 @@
+# Copyright <2016> Akretion
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
```
hparfr, line 2: +2019

hparfr (changes requested):

data_encryption/models/encrypted_data.py
florian-dacosta: There is no need, the name given must be unique, independently of the company.
There is an example in the other module data_encryption_server_environment, the name given is
the name of the model + the database id, se we ensure this name is unique and we can find our encrypted datas easily

data_encryption/__manifest__.py
bealdav: done

data_encryption/models/encrypted_data.py
bealdav: done

data_encryption/__manifest__.py
bealdav: done

data_encryption/__manifest__.py
```diff
@@ -0,0 +1,23 @@
+# Copyright <2019> Akretion
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+{
+    "name": "Encryption data",
+    "summary": "Store accounts and credentials encrypted by environment",
+    "version": "12.0.0.0.1",
+    "category": "Uncategorized",
```
bealdav:
```suggestion
    "category": "Tools",
```

bealdav (approved):

bealdav:
@hparfr could you complete your review ?

data_encryption/__manifest__.py
```diff
@@ -0,0 +1,23 @@
+# Copyright <2019> Akretion
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+{
+    "name": "Encryption data",
+    "summary": "Store accounts and credentials encrypted by environment",
+    "version": "12.0.0.0.1",
```
simahawk: I'd use 1.0.0 + alpha dev status

data_encryption/models/encrypted_data.py
```diff
@@ -0,0 +1,130 @@
...
+
+    _sql_constraints = [
+        ('name_environment_uniq', 'unique (name, environment)',
+         'You can not store multiple encrypted data for the same record and \
+          environment')
+    ]
+
+    @api.model
+    @ormcache('self._uid', 'name', 'env')
+    def _get(self, name, env=None):
```
simahawk: can we give such methods better meaningful names? Eg: `_encrypted_get` `_encrypted_get_json` `_encrypted_store` etc?

data_encryption/models/encrypted_data.py
```diff
@@ -0,0 +1,130 @@
...
+            return cipher.decrypt(encrypted_rec.encrypted_data).decode()
+        except InvalidToken:
+            raise ValidationError(_(
+                "Password has been encrypted with a different "
+                "key. Unless you can recover the previous key, "
+                "this password is unreadable."))
+
+    @api.model
+    @ormcache('self._uid', 'name', 'env')
+    def _get_json(self, name, env=None):
```
simahawk: hmm here you don't get json, you read content that was json and return python data.

data_encryption/models/encrypted_data.py
```diff
@@ -0,0 +1,130 @@
...
+         'You can not store multiple encrypted data for the same record and \
+          environment')
+    ]
+
+    @api.model
+    @ormcache('self._uid', 'name', 'env')
+    def _get(self, name, env=None):
+        if self.env.context.get('bin_size'):
+            self = self.with_context(bin_size=False)
+        if self.env.user.id != SUPERUSER_ID:
```
simahawk: better to use `self.env.user._is_superuser()`

data_encryption/models/encrypted_data.py
```diff
@@ -0,0 +1,130 @@
...
+
+    @classmethod
+    def _get_cipher(cls, env):
+        """Return a cipher using the key of environment.
+        force_env = name of the env key.
+        Useful for encoding against one precise env
+        """
+        key_name = 'encryption_key_%s' % env
+        key_str = config.get(key_name)
+        if not key_str:
```
simahawk, line 99: I don't see any reference in README files about this. Am I wrong?

data_encryption/models/encrypted_data.py
```diff
@@ -0,0 +1,130 @@
...
+        return Fernet(key)
+
+    @api.model
+    def _store(self, name, data,  env=None):
+        if self.env.user.id != SUPERUSER_ID:
+            raise AccessError(
+                _("You can only encrypt data as superuser"))
+        if not env:
+            env = self._retrieve_env()
+        cipher = self._get_cipher(env)
```
simahawk: I would split encrypt/decrypt actions to specific methods so that you can easily replace the way it's done. Eg: `_encrypt_data`, `_decrypt_data`.
Not mandatory tho.

data_encryption/readme/ROADMAP.rst
```diff
@@ -0,0 +1,5 @@
+For now the encryption is dependent on the environment. It has been designed
+to store the same kind of data with different values depending on the environement
+(dev, preprod, prod...).
+An improvement could be to split this in 2 modules. But the environment stuff
+is not a big constraint.
```
simahawk, line 5: Question: what happens if you grab data from PROD into your DEV env? Data will be unreadable, right?
How would you deal w/ it?

simahawk (changes requested):

data_encryption/__manifest__.py
florian-dacosta: You are right, it could be nice to use alpha status for now. Will do

data_encryption/models/encrypted_data.py
florian-dacosta: I don't really see the point as you are inside the EncryptedData class, so it seems clear that the _get will read/decrypt the data ans the _store will write/encrypt data... no?

data_encryption/models/encrypted_data.py
florian-dacosta: Will do

data_encryption/models/encrypted_data.py
florian-dacosta: mmh, have you a method name suggestion then?

data_encryption/models/encrypted_data.py
florian-dacosta: Mmmh, you are right, I was sure there was a mention about this, I'll add it, in the configure section

data_encryption/readme/ROADMAP.rst
florian-dacosta: I am not sure I understand your question.
To resume (I'll mention this in the configure part of the readme) you have one "encryption key" by environment.
So, in your dev environment, if you try to read a prod data, it will fail and raise an error, because the prod encryption key won't be present. But, it should not be possible.
I mean, this module is really low level and is meant to be used by other (example with server_environment_data_encryption) but these module should not let a dev environment even try to read a prod data... It should not be possible, and if it is, it will raise an error anyway.

The idea would be to configure all encryption key in your prod environment, and only the environment key for others.
For instance, in your dev config file, you will have 
encryption_key_dev = XXX
running_env=dev

In your prod environment, you will have : 
encryption_key_dev = XXX
encryption_key_prod=YYY
running_env=prod

This allows you to set the values for all your environment directly from the production. (there is a was to change environment to set read the values in server_environment_data_encryption)
And indeed, in your dev environment, Odoo won't be able to decrypt the data.
But in principle, no module should make it possible.

Not sure I answered you, do not hesitate to precise your question if I was not clear!

florian-dacosta:
@simahawk Thanks for your review, I have attended most of them.
Please, give me a feedback about the configure part of the readme.
@hparfr Still need you to complete your review please!

data_encryption/models/encrypted_data.py
simahawk: @florian-dacosta because it makes easier to understand what it does, especially if you inspect model's methods via python autocomplete/debugger.

data_encryption/models/encrypted_data.py
simahawk: `_encrypted_read_json`?

data_encryption/readme/ROADMAP.rst
simahawk: ok, thanks for the explanation and the README update :)

data_encryption/security/ir.model.access.csv
```diff
@@ -0,0 +1,2 @@
+id,name,model_id:id,group_id:id,perm_read,perm_write,perm_create,perm_unlink
+access_encrypted_data,access_encrypted_data,model_encrypted_data,base.group_system,0,0,0,0
```
simahawk, line 2: system users cannot read them?

florian-dacosta: No. This table is meant to contain sensitive and diverse data.
So probably no user should be able to access the whole table. 

When you implement a submodule, you should call the _get/_store methods with sudo.
So it is the rôle of each submodule to implement who is able to read or store data, but by default, nobody can except the superuser.
So you could have a module that allow to store bank passwords for instance
Another carrier password, and a system user may be able to access only the carrier stuff but not the bank stuff..

data_encryption/models/encrypted_data.py
florian-dacosta: @simahawk Well I am not really convinced about this one, even if you inspect, you know on which model you are and 1_get/_store seems clean enough, adding _encrypted seems redondant. But anyway, I just did it by courtesy, I don't really mind, so let's do this as there is not yet module dependent on it yet!

florian-dacosta:
@simahawk Comments attended!

data_encryption/models/encrypted_data.py
```diff
@@ -0,0 +1,138 @@
...
+        """Return a cipher using the key of environment.
+        force_env = name of the env key.
+        Useful for encoding against one precise env
+        """
+        key_name = 'encryption_key_%s' % env
+        key_str = config.get(key_name)
+        if not key_str:
+            raise ValidationError(_(
+                "No '%s' entry found in config file. "
+                "Use a key similar to: %s" % (key_name, Fernet.generate_key())
```
gurneyalex: you should not have the parameter substitution in the translated string: please move the `%` outside the call to `_()`

gurneyalex (approved):
one minor change requested.

data_encryption/models/encrypted_data.py
florian-dacosta: @gurneyalex Good catch!
I fixed it, thanks

simahawk (approved):

simahawk:
@florian-dacosta would be nice to squash your commits into `[12.0] Add data_encryption` and add details regarding origin from keychain into  extra message.

florian-dacosta:
I think this one is ready to merge now, not sure why gitbot does not say so!
@sebastienbeau You can review/merge this ?

sebastienbeau (approved):
/ocabot merge

dreispt:
Hello, I'm considering promoting this module to Mature level.
In your opinion, what should be done toward that?

Thanks

florian-dacosta:
I am not sure, this module is quite simple and probably could be mature, but it is kind of a base module, as it does not do anything by itself. It implements methods to encrypt and decrypt data, but it needs other modules to encrypt or decrypt something.

AFAIK, the only module implementing a feature with it is `server_environment_data_encryption`.
This one is not mature, mainly because we have no feedback of use on a production server, so it is hard to put it as mature for now.

dreispt:
Thank you @florian-dacosta 

Simple is good.
I need an encrypted field implementation, and using this as a base for that fits the requirements (having the key on the conf file is acceptable).
I made a crude Poc for that -> https://github.com/OCA/server-env/pull/50

If customer approves, I will develop that into a proper implementation, probably adding an `encrypted=True` attribute to Fields.
Having the base `data_encryption` in alpha/beta state is an issue though, so I would like to promote it.

hparfr:
> This one is not mature, mainly because we have no feedback of use on a production server, so it is hard to put it as mature for now.

I have one customer in prod with this module since few weeks. No problem so far.

IMO it should be "stable" instead of "mature".

dreispt:
@hparfr Thanks for the information. Can you share the use case? server_environment secrets?

florian-dacosta:
@dreispt 
Yes, `server_environment` module allow to configure some environment dependent variables in odoo configuration files, or in database but not encrypted.

The secrets are safe in configuration file, but the odoo users can't configure it themselve.
If they store the data in the database, it is not safe.

The goal of the module is to allow to store the environement dependent variables in the database, but encrypted.

So an odoo user, can configure its environment dependent data himself, in a secure way.
Use cases for this could be carrier account, external storage accounts (ftp, etc...) where we want the user to be able to configure it himself, but we want to protect the passwords.

And it is fully compatible with server_environment module, you still can store the data in the odoo configuration files.

hparfr:
> @hparfr Thanks for the information. Can you share the use case? server_environment secrets?

In this project (e-commerce), the following modules have different keys per environnment:
- mail_environment 
- storage (for OpenStack Object Storage)
- search_engine (elastic search)
- shopinvader locomotive

The goals are:
- credentials in database and secure at rest
- a backup of a database can be used directly in CI or in dev - without cleaning credentials

PR #24, opened 2019-05-14
[12.0] Add module server_environment_data_encryption

florian-dacosta:
This module aims to replace keychain.
At the start keychain came from the need to manage different value for different environments (like server_environment) but making the configuration possible by the user (directly in the UI and not in configuration file.) The second need was to have encrypted data, for example, to avoid giving critical information hen you give a backup for developement purpose.

The main issue is that not everyone want to use keychain (Server_environement is used a lot in the community) and so it can't be a dependencie in some base module, which forces to do a lot of 'glue' module each time we need it.

Since the recent great improvements made in server_environment, we can now do something compatible.
So this module depends on server_environment and allow to configure data by environements, directly in the User Interface and encrypted in the database.

First, odoo will try to get the values from server_environement, if it gets nothing, it will fallback on the encrypted data in the database.
Also, it allows user to configure the data for all the environement from the place.
For instance, if you install mail_environment and this module, from the incoming mail server form, you have smart buttons to switch to dev environment and set values, then for preproduction environment and set the values, and the production environement, of course.
So it is really easy for a normal user to make these configuration if necessary.

Depends on https://github.com/OCA/server-env/pull/23

@sebastienbeau @hparfr @guewen @lmignon

### DIFF
--- a/server_environment_data_encryption/__manifest__.py
+++ b/server_environment_data_encryption/__manifest__.py
@@ -4,11 +4,9 @@
     "version": "12.0.1.0.0",
     "development_status": 'Alpha',
     "category": "Tools",
-    "website": "https://akretion.com/",
+    "website": "https://github.com/OCA/server-env",
     "author": "Akretion, Odoo Community Association (OCA)",
     "license": "AGPL-3",
-    "application": False,
     "installable": True,
     "depends": ["server_environment", "data_encryption"],
-    "data": [],
 }
--- a/server_environment_data_encryption/__manifest__.py
+++ b/server_environment_data_encryption/__manifest__.py
@@ -1,0 +1,13 @@
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+{
+    "name": "Server Environment Data Encryption",
+    "version": "12.0.0.0.1",
+    "category": "Tools",
+    "website": "https://akretion.com/",
+    "author": "Akretion, Odoo Community Association (OCA)",
+    "license": "AGPL-3",
+    "application": False,
+    "installable": True,
+    "depends": ["server_environment", "data_encryption"],
+    "data": [],
+}
--- a/server_environment_data_encryption/__manifest__.py
+++ b/server_environment_data_encryption/__manifest__.py
@@ -1,7 +1,8 @@
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 {
     "name": "Server Environment Data Encryption",
-    "version": "12.0.0.0.1",
+    "version": "12.0.1.0.0",
+    "development_status": 'Alpha',
     "category": "Tools",
     "website": "https://akretion.com/",
     "author": "Akretion, Odoo Community Association (OCA)",
--- a/server_environment_data_encryption/models/server_env_mixin.py
+++ b/server_environment_data_encryption/models/server_env_mixin.py
@@ -118,13 +118,17 @@ class ServerEnvMixin(models.AbstractModel):
         current_env = self.env.context.get("environment") or config.get(
             "running_env"
         )
-        other_environments = [
+        # Important to keep this list sorted. It makes sure the button to
+        # switch environment will always be in the same order. (more user
+        # friendly) and the test would fail without it as the order could
+        # change randomly and the view would then also change randomly
+        other_environments = sorted([
             key[15:]
             for key, val in config.options.items()
             if key.startswith("encryption_key_")
             and val
             and key[15:] != current_env
-        ]
+        ])
 
         if not current_env:
             raise ValidationError(
--- a/server_environment_data_encryption/models/server_env_mixin.py
+++ b/server_environment_data_encryption/models/server_env_mixin.py
@@ -1,0 +1,165 @@
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+import logging
+from odoo import api, models, _
+from odoo.exceptions import ValidationError
+from odoo.tools.config import config
+from lxml import etree
+from odoo.osv.orm import setup_modifiers
+
+
+_logger = logging.getLogger(__name__)
+
+
+class ServerEnvMixin(models.AbstractModel):
+    _inherit = "server.env.mixin"
+
+    def _compute_server_env_from_default(self, field_name, options):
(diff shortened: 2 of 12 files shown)


rvalyi:
Hello @florian-dacosta 
I would advocate for using LGPL3 here. I think the LGPL part of Odoo is getting large (we can like it or not but that's a fact). But putting such base infrastructure as AGPL, I think the only thing we do is limiting its adoption and making it likely a competing LGPL solution to appear and take over this module no matter how good it is. We would fool ourselves thinking people would suddenly open their codebase instead of rewritting such a module. So personally I think it is still good that the OCA keep fighting for AGPL on more sophisticated ERP modules for which openness is the major selling point but for basic infrastructure modules like this one that could be appealing for startups using Odoo as a framework I think we better think twice. BTW I think the same about base_jsonify and a few other Akretion framework modules.

sebastienbeau:
Hi on my side I always prefers to use AGPLv3 first and then see if somebody ask for changing it.
Moreover server env is not LGPL so AGPL is good

gurneyalex (approved):

server_environment_data_encryption/__manifest__.py
```diff
@@ -0,0 +1,14 @@
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+{
+    "name": "Server Environment Data Encryption",
+    "version": "12.0.1.0.0",
+    "development_status": 'Alpha',
+    "category": "Tools",
+    "website": "https://akretion.com/",
```
simahawk: https://github.com/OCA/server-env

server_environment_data_encryption/__manifest__.py
```diff
@@ -0,0 +1,14 @@
...
+    "author": "Akretion, Odoo Community Association (OCA)",
+    "license": "AGPL-3",
+    "application": False,
```
simahawk: no need for application flag

server_environment_data_encryption/__manifest__.py
```diff
@@ -0,0 +1,14 @@
...
+    "installable": True,
+    "depends": ["server_environment", "data_encryption"],
+    "data": [],
```
simahawk: no need for data

server_environment_data_encryption/models/server_env_mixin.py
```diff
@@ -0,0 +1,169 @@
...
+
+
+_logger = logging.getLogger(__name__)
+
+
+class ServerEnvMixin(models.AbstractModel):
+    _inherit = "server.env.mixin"
+
+    def _compute_server_env_from_default(self, field_name, options):
+        "First return database encrypted value then default value"
```
simahawk: triple quotes

server_environment_data_encryption/models/server_env_mixin.py
```diff
@@ -0,0 +1,169 @@
...
+                values.update(new_val)
+                encrypted_data_obj._encrypted_store_json(
+                    encrypted_data_name, values, env=env
+                )
+
+    def action_change_env_data_encrypted_fields(self):
+        action_id = self.env.context.get("params", {}).get("action")
+        if not action_id:
+            # We don't know which action we are using... take one random.
+            action_id = self.env['ir.actions.act_window'].search(
```
simahawk: you can use `self.get_formview_action`

server_environment_data_encryption/models/server_env_mixin.py
```diff
@@ -0,0 +1,169 @@
...
+        action["res_id"] = self.id
+        views_form = []
+        for view_id, view_type in action.get("views", []):
+            if view_type == "form":
+                views_form.append((view_id, view_type))
+        action["views"] = views_form
+        return action
+
+    def _get_extra_environment_info_div(self, current_env, extra_envs):
+        button_div = "<div>"
```
simahawk, line 73: would nicer and easier to use a qweb template to generate this.
Not a blocker but worth a TODO comment.

server_environment_data_encryption/tests/test_server_environment_data_encrypt.py
```diff
@@ -0,0 +1,36 @@
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from odoo.addons.data_encryption.tests.common import CommonDataEncrypted
+from pathlib import Path
+
+
+class TestServerEnvDataEncrypted(CommonDataEncrypted):
+
+    #    def test_store_data_no_superuser(self):
```
simahawk: needed?

simahawk (changes requested):

server_environment_data_encryption/__manifest__.py
florian-dacosta: Done

server_environment_data_encryption/__manifest__.py
florian-dacosta: Done

server_environment_data_encryption/__manifest__.py
florian-dacosta: Done

server_environment_data_encryption/models/server_env_mixin.py
florian-dacosta: Done

server_environment_data_encryption/models/server_env_mixin.py
florian-dacosta: I didn't know this one, thanks for the pointer, I made the change.

server_environment_data_encryption/tests/test_server_environment_data_encrypt.py
florian-dacosta: no, I removed it.

server_environment_data_encryption/models/server_env_mixin.py
florian-dacosta: I put a TODO comment, I can't do this now

simahawk (approved):
builds are :red_circle:

florian-dacosta:
Yes, the module data_encryption is not merged yet and it is a dependency

sebastienbeau:
I have rebased the PR so test should be green and ready to merge

sebastienbeau:
let's merge it \o/ /ocabot merge

PR #26, opened 2019-07-19
[IMP] server-env add many2x support

sebastienbeau:


simahawk (commented):
curious: how do you provide values for rel fields?

PR #34, opened 2019-10-23
[12.0][FIX][carrier_environment] remove name from server env fields

florian-dacosta:
Remove name from server env fields as this should not change depending on the environment, since the name is actually used to retrieve the environment related data from the config file

@bealdav @angelmoya @gurneyalex 
Could you do a quick review please?


bealdav (approved):

sebastienbeau (approved):
LGTM

PR #33, opened 2019-10-12
pre-commit, black, isort

sbidoul:
This one requires a manual intervention to fix a pylint error (missing super).

PR #36, opened 2019-12-16
[13.0] pre-commit, black, isort

yvaucher:
take over of https://github.com/OCA/server-env/pull/33


yvaucher:
Let's retry as error was on aptitude

The command "sudo -E apt-get -yq --no-install-suggests --no-install-recommends $(travis_apt_get_options) install expect-dev" failed and exited with 100 during .

yvaucher:
just had to insist

PR #37, opened 2019-12-19
[13.0][FIX] server_environment: Change import for Serialized field

rousseldenis:
Can cause errors like :

image


rousseldenis:
@guewen @lmignon

guewen (approved):
Thanks

lmignon (approved):

PR #42, opened 2020-03-09
[FIX] server_environment: Change import for Serialized field

simahawk:
Backport of https://github.com/OCA/server-env/commit/a101ef181d2039b51f15576dfe2eeba9a7befd37

Companion of https://github.com/OCA/server-env/pull/41


simahawk:
fast tracking as it breaks builds on related repos w/ oca_dependencies propagation

PR #45, opened 2020-04-01
server_environment: running_env default to `test` (fix #44)

### LINKED ISSUE #44: Odoo Critical error "failed to initialize database" once server_environemnt is installed
```
server-env/server_environment/serv_config.py", line 50, in <module>
    "The parameter 'running_env' has not be set neither in base config "
Exception: The parameter 'running_env' has not be set neither in base config file option -c or in openerprc.
We strongly recommend against using the rc file but instead use an explicit config file with this content:
[options]
running_env = dev
```

I get the point that the additional configuration is very important.
But does it justify preventing the service from starting?
I think this would be a log message, rather than a error with a traceback.

simahawk:


bealdav (approved):

simahawk:
tests to be fixed. No time now, I'll get back ASAP.

.travis.yml
```diff
@@ -41,8 +41,6 @@ install:
     ${HOME}/maintainer-quality-tools
   - export PATH=${HOME}/maintainer-quality-tools/travis:${PATH}
   - travis_install_nightly
-  # Requirements to test server_environment modules
-  - printf '[options]\n\nrunning_env = dev\n' > ${HOME}/.openerp_serverrc
```
dreispt, line 45: 🎉

dreispt (approved):

dreispt:
Tests need some love.

server_environment/server_env.py
```diff
@@ -184,9 +190,6 @@ def _build_model(cls, pool, cr):
         """
         ModelClass = super(ServerConfiguration, cls)._build_model(pool, cr)
         ModelClass._add_columns()
-        ModelClass.running_env = system_base_config["running_env"]
```
simahawk, line 187: this seems not required as is not used anywhere

bodedra (approved):
LGTM :+1:

pedrobaeza (approved):
I still don't understand why this is "activated" by default without installing the module, but if working, OK for me.

I wish also that you use Doodba for deploying avoiding all these modules, but that's too much, hehe.

simahawk:
I squash and merge as it gets green again. Runbot is :orange_circle: only indeed.

simahawk:
@pedrobaeza I had to squash... :disappointed:

pedrobaeza:
Oops, too late... I have just merged directly by the same reason. OCA runbot needs an urgent fix. Maybe unsticky more branches.

pedrobaeza:
Ouch! I always squash on the fly for avoiding extra CIs. Sorry for being so impatient and not look to that. I just want to have green CIs ASAP :angel:

simahawk:
> I wish also that you use Doodba for deploying avoiding all these modules, but that's too much, hehe.

why would it help?

pedrobaeza:
You won't need these modules with that deployment method.

Note that Doodba serves for production, staging and devel environments.

simahawk:
eh same here w/ our docker image. The point is: how do you deal w/ model / modules / params / whatever specific configurable variables?
This is not about odoo config only.

pedrobaeza:
We have our scaffolding for that, that now has been powered-up with copier templating system. AFAIK, you have modules for:

- mail configuration. We put it on the scaffolding.
- different Odoo configurations. The same.
- ir_config_parameter. We inject this through Ansible, as we consider this more a DevOps, but scaffolding can be expanded for that if you switch and consider it interesting :wink: 

Note that secrets are not hardcoded in the scaffolding.

simahawk:
we do the same, we have templates for them.
The point is, if you want to customize via env the behavior of a model you cannot do it w/ templating. 
For instance, the mail server: https://github.com/OCA/server-env/blob/13.0/mail_environment/models/ir_mail_server.py#L12
And for secrets: this is already handled as well. You can have public vars and secret vars.

pedrobaeza:
Uhm, having Mailhog without configuring anything in our Doodba staging seems enough for us sincerely.

dreispt:
Thank you!
Can this be backported to 12.0 please?

simahawk:
v11 https://github.com/OCA/server-env/pull/46

alexey-pelykh:
v12 https://github.com/OCA/server-env/pull/47

PR #49, opened 2020-05-20
[12.0][PORT] server_environment. port static files controllers (from server-tools 8.0 branch)

legalsylvain:
This PR is the port of the PR against OCA / server-tools 8.0 branch. ([ref](https://github.com/OCA/server-tools/pull/1503))

It improves the module ``server_environment`` to have the possibility to use custom static files, depending on the environment.

# Usage

* Create a file view/templates.xml, and insert a css file, for exemple

```
    <template id="login_layout_no_db" name="Login Layout" inherit_id="web.login_layout" >
        <xpath expr="." position="inside">
            <link rel="stylesheet" href="/server_environment_files_sample/static/RUNNING_ENV/web_login.css"/>
        </xpath>
    </template>
```

Then create your custom css files in the according folders, in your ``server_environment_files`` folders

```

    /server_environment_files/static/dev/css.css
    /server_environment_files/static/prod/css.css
    ....

```

# Technical information

this module adds basically a new http controller, that catch dynamically the call of the URLs that match the pattern ``server_environment_files/static/RUNNING_ENV/<path>`` and make a redirection. 

# Sample

image

CC : Original reviewers : @fcayre, @quentinDupont

### DIFF
--- a/server_environment/controllers/main.py
+++ b/server_environment/controllers/main.py
@@ -1,0 +1,30 @@
+# Copyright (C) 2019 - Today: GRAP (http://www.grap.coop)
+# @author: Sylvain LE GAL (https://twitter.com/legalsylvain)
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from werkzeug.utils import redirect
+from urllib.parse import urljoin
+
+from odoo.tools.config import config
+from odoo import http
+from odoo.http import request, Controller
+
+
+class ServerEnvironmentController(Controller):
+
+    @http.route(
+        '/server_environment_<module_extension>'
+        '/static/RUNNING_ENV/<local_path>',
+        type='http', auth='public')
+    def environment_redirect(self, module_extension, local_path, **kw):
+        # Note: module_extension is present to make working
+        # the module in normal configuration, with the folder
+        # server_environment_files and in demo configuration, with the
+        # module  server_environment_files_sample
+        running_env = config.get('running_env', "default")
+        IrConfigParameter = request.env['ir.config_parameter'].sudo()
+        base_url = IrConfigParameter.get_param('web.base.url', '')
+        new_path = "/server_environment_%s/static/%s/%s" % (
+            module_extension, running_env, local_path)
+        url = urljoin(base_url, new_path)
+        return redirect(url, 303)
--- a/server_environment/__init__.py
+++ b/server_environment/__init__.py
@@ -18,6 +18,7 @@
 #
 ##############################################################################
 from . import models
+from . import controllers
 # TODO when migrating to 12, fix the import of serv_config by renaming the
 # file?
 # Add an alias to access to the 'serv_config' module as it is shadowed
--- a/server_environment/__manifest__.py
+++ b/server_environment/__manifest__.py
@@ -9,7 +9,7 @@
         "base",
         "base_sparse_field",
     ],
-    "author": "Camptocamp,Odoo Community Association (OCA)",
+    "author": "Camptocamp,GRAP,Odoo Community Association (OCA)",
     "summary": "move some configurations out of the database",
     "website": "http://github.com/OCA/server-env",
     "license": "GPL-3 or any later version",
--- a/server_environment/controllers/__init__.py
+++ b/server_environment/controllers/__init__.py
@@ -1,0 +1,1 @@
+from . import main
--- a/server_environment/readme/CONFIGURE.rst
+++ b/server_environment/readme/CONFIGURE.rst
@@ -99,7 +99,9 @@ This module provides also the possibility to load static files depending
 on the environment.
 
 Create a file view/templates.xml, and insert a css file
-::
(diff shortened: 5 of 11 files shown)


server_environment/controllers/main.py
```diff
@@ -0,0 +1,30 @@
...
+from odoo.tools.config import config
+from odoo import http
+from odoo.http import request, Controller
+
+
+class ServerEnvironmentController(Controller):
+
+    @http.route(
+        '/server_environment_<module_extension>'
+        '/static/RUNNING_ENV/<local_path>',
```
simahawk, line 17: Is this `RUNNING_ENV` bit useful? Seems misleading to me :thinking: 
Since the controller is responsible for the lookup it's pointless IMO.

server_environment/controllers/main.py
```diff
@@ -0,0 +1,30 @@
...
+        type='http', auth='public')
+    def environment_redirect(self, module_extension, local_path, **kw):
+        # Note: module_extension is present to make working
+        # the module in normal configuration, with the folder
+        # server_environment_files and in demo configuration, with the
+        # module  server_environment_files_sample
+        running_env = config.get('running_env', "default")
+        IrConfigParameter = request.env['ir.config_parameter'].sudo()
+        base_url = IrConfigParameter.get_param('web.base.url', '')
+        new_path = "/server_environment_%s/static/%s/%s" % (
```
simahawk, line 27: if the env has no file this will give 404 every time you hit it.
Imagine you want to change styles only to make prominent that you are on test instead of prod.
On prod you'll get tons of 404, unless you recommend to always have a file for all envs (?).

legalsylvain: >  On prod you'll get tons of 404, unless you recommend to always have a file for all envs (?).

Exactly. As for the other part of this module. you should have a folder for each environment, event if it's empty.

server_environment/controllers/main.py
legalsylvain: No it's important. The call will be done first to the URL with RUNNING_ENV, then after the 304, with the correct environment.

server_environment/controllers/main.py
legalsylvain: Note : it is written in the ``CONFIGURE.rst`` file : 

```
Then, create css files for each environment you have. exemple:


    /server_environment_files/static/dev/css.css
    /server_environment_files/static/prod/css.css
    ...
```

server_environment/controllers/main.py
simahawk: Yeah, I mean that even if you remove "RUNNIG_ENV" you'll get this call and nothing would change in this behavior ;)
Anyway, not a big deal but I founded it misleading because people may expect to replace that capital letter bit w/ the env in their xml declaration.

OTOH there might be another way to achieve the same result.
Just speaking out loud, here is it: 

* remove the controller
* rely on a ir config param that can be already configured by env (w/ [this module](https://github.com/OCA/server-env/tree/13.0/server_environment_ir_config_parameter))
* look up for it in the template: render the css link only if the param is found

You'll get less code to maintain and you can turn on/off the feature as you like.
No force need for an extra file you might not need, no 404.

Eg:

```xml
 <template id="login_layout_no_db" name="Login Layout" inherit_id="web.login_layout" >
        <xpath expr="." position="inside">
            <t t-set="server_env_css_path" t-value="request.env['ir.config_param"].sudo().get_param("server_env_css_path", None)" />
            <link t-if="server_env_css_path" rel="stylesheet" t-att-href="server_env_css_path"/>
        </xpath>
    </template>
```
And path will be something like `/server_environment_files/prod/styles.css`. 
Or from another module as you'll the freedom.

Not tested, it should work, in theory :wink: 
Oh, then the module seems not even needed with this approach :thinking: 

What do you think?

legalsylvain: > What do you think?

I love to participate in the OCA. I've been working on Odoo for 8 years, and I'm always learning new things, and new ways of looking at things.
You're absolutely right. You come up with a better design, with less code to maintain.

legalsylvain:
Closing after https://github.com/OCA/server-env/pull/49#discussion_r429062513

server_environment/controllers/main.py
simahawk: > I love to participate in the OCA. I've been working on Odoo for 8 years, and I'm always learning new things, and new ways of looking at things.

Same here, I love open source baby! :smile: 

>  You're absolutely right. You come up with a better design, with less code to maintain.

Well, we still need a PoC to see if it works or not :wink:  
If you try it, please, let me know if it works properly.
Maybe I'm wrong and I'll learn something else as well :nerd_face:

legalsylvain: > Well, we still need a PoC to see if it works or not wink
If you try it, please, let me know if it works properly.

It works perfectly well.

https://github.com/grap/grap-odoo-custom/tree/12.0/server_environment_files

PR #52, opened 2020-07-15
Add mixin for tech name

simahawk:
Moving here something done on search engine https://github.com/OCA/search-engine/blob/12.0/connector_search_engine/models/se_backend.py#L43

It allows to have more reliable naming for env reference.

### DIFF
--- a/server_environment/models/server_env_tech_name_mixin.py
+++ b/server_environment/models/server_env_tech_name_mixin.py
@@ -1,0 +1,72 @@
+# Copyright 2020 Camptocamp (http://www.camptocamp.com)
+# @author Simone Orsi <simahawk@gmail.com>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import api, fields, models
+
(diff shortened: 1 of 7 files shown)


i-vyshnevska (approved):

server_environment/models/server_env_tech_name_mixin.py
```diff
@@ -0,0 +1,71 @@
...
+    _name = "server.env.techname.mixin"
+    _description = "Server environment technical name"
+    _sql_constraints = [
+        ("tech_name_uniq", "unique(tech_name)", "`tech_name` must be unique!",)
+    ]
+
+    tech_name = fields.Char(
+        required=True,
+        help="Unique name for technical purposes. " "Eg: server env keys.",
+    )
```
guewen, line 39: this field seems to be a good candidate for using the new editable and stored computed field, that would remove the need for the onchange, create, write...

simahawk: added to ROADMAP :wink:

guewen (approved):

PR #53, opened 2020-07-24
[13.0] Create module server_environment_delivery

grindtildeath:
This module allows to configure shipping methods according to server environment.

At the moment, the module only allows to define the field prod_environment by
defining a `[delivery_carrier]` with `prod_environment` key as follows:

Restrict usage of prod environment:

  ```
  [delivery_carrier]
  prod_environment=False
  ```

Force usage of prod environment:

  ```
  [delivery_carrier]
  prod_environment=False
  ```

If the key is not set, the user will still be able to switch the value.

### DIFF
--- a/server_environment_delivery/models/delivery_carrier.py
+++ b/server_environment_delivery/models/delivery_carrier.py
@@ -1,0 +1,15 @@
+# Copyright 2020 Camptocamp SA
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl)
+from odoo import models
+
+
+class DeliveryCarrier(models.Model):
+    _name = "delivery.carrier"
+    _inherit = ["delivery.carrier", "server.env.mixin"]
+
+    @property
+    def _server_env_fields(self):
+        base_fields = super()._server_env_fields
+        delivery_fields = {"prod_environment": {}}
+        delivery_fields.update(base_fields)
+        return delivery_fields
--- a/server_environment_delivery/models/delivery_carrier.py
+++ b/server_environment_delivery/models/delivery_carrier.py
@@ -5,7 +5,7 @@ from odoo import models
 
 class DeliveryCarrier(models.Model):
     _name = "delivery.carrier"
-    _inherit = ["delivery.carrier", "server.env.mixin"]
+    _inherit = ["delivery.carrier", "server.env.techname.mixin", "server.env.mixin"]
 
     @property
     def _server_env_fields(self):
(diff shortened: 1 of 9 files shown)


simahawk:
I think this should go to OCA/delivery instead. This repo has only basic env modules.

server_environment_delivery/models/delivery_carrier.py
```diff
@@ -0,0 +1,15 @@
+# Copyright 2020 Camptocamp SA
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl)
+from odoo import models
+
+
+class DeliveryCarrier(models.Model):
+    _name = "delivery.carrier"
+    _inherit = ["delivery.carrier", "server.env.mixin"]
```
simahawk: I'd recommend to use the brand new mixin `server.env.techname.mixin` because the carrier name could change based on whatever needs.

grindtildeath:
@simahawk Sorry but this is the right repository for this module as it doesn't bring any functionality in terms of delivery but only adds management of a field through server environment... 

Moreover, considering this https://github.com/OCA/stock-logistics-warehouse/pull/942#issuecomment-659338322 I don't see any reason to do the change :wink:

guewen:
I have always put the 'server_env' glue in the same repository where the module, not here. I think it's easier to keep track of them on migrations if they are alongside their companion module.

simahawk:
> @simahawk Sorry but this is the right repository for this module as it doesn't bring any functionality in terms of delivery but only adds management of a field through server environment...

No, you are adding a specific module here where we have only generic modules.
And this repo will have to be dependent on delivery.

> Moreover, considering this [OCA/stock-logistics-warehouse#942 (comment)](https://github.com/OCA/stock-logistics-warehouse/pull/942#issuecomment-659338322) I don't see any reason to do the change wink

Oh c'on, that's not fair :smile: 

1. I didn't say we should not move it
2. the module was already in use and already released in the same repo
3. here's a brand new module which is not used yet and does not need renaming
4. if I did it wrong... do not copy me :stuck_out_tongue_winking_eye:

grindtildeath:
In terms of dependencies, `delivery` module is in odoo/odoo and the only other dependency is `server_environment` which is a module hosted in this OCA/server-env repository. So I really don't see alongside which other module should this one be hosted...

> No, you are adding a specific module here where we have only generic modules.
> And this repo will have to be dependent on delivery.

Please explain:
- The difference between specific and generic in this particular case
- Why should this repo depend on OCA/delivery-carrier because the only dependency on this PR is on odoo/odoo

simahawk:
moved here https://github.com/OCA/delivery-carrier/pull/351

PR #61, opened 2020-10-09
[13.0] [MIG] server_environment_data_encryption

ThomasBinsfeld:


data_encryption/models/encrypted_data.py
```diff
@@ -0,0 +1,139 @@
...
+                )
+            )
+
+    @api.model
+    @ormcache("self._uid", "name", "env")
+    def _encrypted_get(self, name, env=None):
+        if self.env.context.get("bin_size"):
+            self = self.with_context(bin_size=False)
+        if not self.env.user._is_superuser():
+            raise AccessError(_("Encrypted data can only be read as superuser"))
```
ThomasBinsfeld: @florian-dacosta Why the encryption model is meant to be used only as superuser? It's problematic in my project.

florian-dacosta: Hi @ThomasBinsfeld 
The encryption model is meant to contained all encrypted data (from any model).
So, it may containing many different sensible data (passwords...).

To be secure, we restrict its access to the superuser. Indeed, if one use can access this model, then it can access all sensitive data across all models.
Each module submodule implementing the encryption/decryption of data should manage its case using sudo.

It is the case for data_encryption_server_environement, which handle the encryption of data for all server dependent fields.
https://github.com/OCA/server-env/blob/12.0/server_environment_data_encryption/models/server_env_mixin.py#L22
https://github.com/OCA/server-env/blob/12.0/server_environment_data_encryption/models/server_env_mixin.py#L40

If you create another sub-module made to encrypt/decrypt another type of data, I guess you should implement it with sudo also.
I think it would be a security breach to change this


Why is it problematic for your project?

ThomasBinsfeld: Thanks for your response. I think i understand what my problem is.

In v13, `sudo()` doesn't change the current user. So `self.env.user._is_superuser()` is False, even if the method is called with `sudo()`. That's why i have access errors everywhere in my project.

In `data_encryption`, i need to replace `if not self.env.user._is_superuser():` by `if not self.env.user._is_superuser() and not self.env.su:` if i want to detect the `sudo()`.

In order to preserve the initial behavior, do you think i should replace the two `sudo()` you pointed me by a `with_user(SUPERUSER_ID)`?

florian-dacosta:
Mmmh, I do not think it is necessary to keep the super user stuff.

Like you said, `sudo` worked differently in v12- and the user was changed to super user, which is not the case anymore.
So I guess in v13, we just have to check `self.env.su` and we do not need anymore to check the user.
I did not work for v13 yet, so I am not so confortable with this and could be mistaken, but I think it is enough.

What do you think?

data_encryption/models/encrypted_data.py
florian-dacosta: Mmmh, I do not think it is necessary to keep the super user stuff.

Like you said, sudo worked differently in v12- and the user was changed to super user, which is not the case anymore.
So I guess in v13, we just have to check self.env.su and we do not need anymore to check the user.
I did not work for v13 yet, so I am not so confortable with this and could be mistaken, but I think it is enough.

What do you think?

ThomasBinsfeld: Done :ok_hand:

.pre-commit-config.yaml
```diff
@@ -9,7 +9,9 @@ exclude: |
   # Repos using Sphinx to generate docs don't need prettying
   ^docs/_templates/.*\.html$|
   # You don't usually want a bot to modify your legal texts
-  (LICENSE.*|COPYING.*)
+  (LICENSE.*|COPYING.*)|
+  # server_environment_data_encryption tests files
+  ^server_environment_data_encryption/tests/fixtures/.*$
```
simahawk, line 14: I'd change this to `*/tests/fixtures/*` and maybe propose it to https://github.com/OCA/oca-addons-repo-template

data_encryption/models/encrypted_data.py
```diff
@@ -0,0 +1,143 @@
...
+    def _encrypted_get(self, name, env=None):
+        if self.env.context.get("bin_size"):
+            self = self.with_context(bin_size=False)
+        if not self.env.su:
+            raise AccessError(
+                _("Encrypted data can only be read with suspended security (sudo)")
+            )
+        if not env:
+            env = self._retrieve_env()
+        encrypted_rec = self.search([("name", "=", name), ("environment", "=", env)])
```
simahawk, line 65: limit=1?

server_environment_data_encryption/readme/CONFIGURE.rst
```diff
@@ -0,0 +1,23 @@
...
+  running_env=preprod
+  encryption_key_preprod=YYY
+
+Production environment ::
+
+  [options]
+  running_env=prod
+  encryption_key_dev=XXX
+  encryption_key_preprod=YYY
+  encryption_key_prod=ZZZ
```
simahawk, line 23: I think there's an important step missing here: how does it works?
How am I supposed to provide secret configuration for my records?

By reading the code I understand that when you install it is going to convert env values to encrypted.data records (one per each record). Correct?

But then, how am I supposed to provide my secrets via env keys?
I mean, can I achieve something like I would do w/ `SERVER_ENV_CONFIG_SECRET`?

My guess: you cannot. You rely on manual configuration of the values via UI and simply store them encrypted.
Which means: at first setup, you'll have to configure manually each param on each env.
Unless you generate encrypted.data records w/ some black magic upgrade step/hook.

It would be nice if someone can document this in the README :pray:

simahawk (changes requested):

data_encryption/models/encrypted_data.py
florian-dacosta: there is a unique sql constraint on `(name, env)`
I guess it is useless to have a limit here

server_environment_data_encryption/readme/CONFIGURE.rst
florian-dacosta: If I understand correctly, you think we should add a post install hook to encrypt all env dependent clear data inf the database ?
If yes, you are right, there are no such thing yet, and it could be a nice addition. Still it may be tricky because nothing prevent you to install the module without having configured the entryption key yet...

So maybe it would be best to propose a shell script or something someone could run manually if needed it its situation.
But anyway, if you install the module on existing database with password in clear, yes they still will be in clear and it does not change anything. Only editing it again will encrypt it.

For  the how does it works, it is quite simple : 
Each time someone write an env dependent fields, instead of storing it a field in clear, Odoo will read all env dependent values already stored/encrytpted, update it with the modified value and re-encrypt it.

I am not sure about you question about secret configuration, it has to be in the odoo config file like stated in the readme.

simahawk: > If I understand correctly, you think we should add a post install hook to encrypt all env dependent clear data inf the database ?

Nope, I think you should mention this and all the other things you mentioned in this answer that today are not very clear in the README. Then anything else (scripts, hooks, whatever) should stay in a ROADMAP section.

> If yes, you are right, there are no such thing yet, and it could be a nice addition. Still it may be tricky because nothing prevent you to install the module without having configured the entryption key yet...

> So maybe it would be best to propose a shell script or something someone could run manually if needed it its situation.
> But anyway, if you install the module on existing database with password in clear, yes they still will be in clear and it does not change anything. Only editing it again will encrypt it.

Ok, this should definitely be written in the README :wink: 

Regarding the script/hook: I'm not talking about an install hook just for this module (which would be useful and better anyway). I'm talking about new configuration to be provided automatically by custom modules. 

If I add a new param that should be encrypted I have no other mean - as of today - to create it straight from my module (or any other tool). Thus, I have to rely only on manual edit via UI and I cannot ship my devs in a secure way ready to be deployed on all envs. Correct?

> Each time someone write an env dependent fields, instead of storing it a field in clear, Odoo will read all env dependent values already stored/encrytpted, update it with the modified value and re-encrypt it.

Just to clarify: this happens for EVERY field even if is not useful/required to be encrypted. Correct?
Another thing to be written in the README :wink: 

> I am not sure about you question about secret configuration, it has to be in the odoo config file like stated in the readme.

Yes, I know. The reason why I mentioned it is because I find it more useful / easier to handle in such case (providing secret vars to an instance). I was just asking if one of the purpose of `server_environment_data_encryption` was to avoid using that secret var and how it should be used in relation to it.

data_encryption/models/encrypted_data.py
simahawk: well it was more about making it explicit and faster but as you have indexes for both fields and the constrain... is not needed at all indeed.

PR #62, opened 2020-10-12
14.0 mig server environment

mourad-ehm:

PR #65, opened 2020-10-14
[14.0][MIG] - migration server_environment

sbejaoui:


florian-dacosta:
Seems like a duplicate with https://github.com/OCA/server-env/pull/62

bealdav:
please see https://github.com/OCA/server-env/pull/62

sbejaoui:
> Seems like a duplicate with #62

thank you for noticing me, it's not motioned in the migration issue that's why I missed it.

Issue #25, opened 2019-07-09
server_environment_ir_config_parameter security purpose


bealdav:
Hi all,

cc @sbidoul @guewen @yvaucher @gurneyalex 

I dived into this module

https://github.com/OCA/server-env/blob/12.0/server_environment_ir_config_parameter

I really appreciate to have a way to combine server_env module without to have an UI to build (then a UI fallback). Thanks to the authors

But I have a question about security or confidentiality of informations

A parameter can be a credential (all that we have in server_env keys), and I ask myself why values contained in server_env files
are always written in ir.config_parameter instead of only used as substitution.

https://github.com/OCA/server-env/blob/12.0/server_environment_ir_config_parameter/models/ir_config_parameter.py#L30

IMO, i thought that UI was for demo purposes not for real production storage informations place.

These values are also in backup, and that may imply to rise questions in gdpr context I suppose.

I think we could have an optional behaviour to prevent to store data in db.
What do you think?

Maybe I missed something, so please make me take the right way.

Thanks

sbidoul:
I wrote this module. The initial goal was to override keys in arbitrary modules that store their config in ir.config.parameter. I think my first use case was for `web_environment_ribbon`.

So in those use cases, this module is not less nor more secure than the original that stores it's config in database.

I chose to store it in database for maximum compatibility. But I would be fine with a substitution approach. To be verified if there are no place that read from ir.config.parameters using another method than get_param.

bealdav:
OK thanks.
Another PR in my todo list, then

legalsylvain:
Hi. @bealdav. Did you worked on that point ? I'm using this module and I face the same problem. 
Also, the UI has a end user behaviour that is not intuitive : 

- We have a production database with some config_parameter written in the odoo.cfg files like ``my_key = my_production_value``
- make a dump, and restore the database for test / debug / spare / developpement purpose 
- launch the database in the test environment.
A) going in the ir.config_parameter view, the user will see the ``my_production_value``
B) when "consuming" with ``get_param()`` function, the value in db will be overwritten. 
C) going in the ir.config_parameter view, the user will see the ``my_test_value``

the A situation should not occures.

My proposal : 

- remove the line that writes [in the db](https://github.com/OCA/server-env/blob/12.0/server_environment_ir_config_parameter/models/ir_config_parameter.py#L30).
- add a field ``environment_value``, computed.
- display the field if set.

Pro:
-> fixes security issues. developer / tester will not have access to the production values;
[...]

PR #67, opened 2020-10-16
[IMP] server_environment_ir_config_parameter: display if parameter are defined by env

legalsylvain:
- display in the config parameter list and form views the new field ``is_environment``, and make the ``key`` and ``value`` readonly if checked.
- remove the roadmap section.

image

kind regards.

PR #68, opened 2020-10-16
[12.0] fix server environmnet ir config parameter security issues

legalsylvain:
Hi all. 

[x] Display database / environment parameters clearly.
[x] avoid to display production values in testing environment.
[x] improve documentation regarding that points.

ir_config_parameter_tree

closes: #25

Waiting for : #67. Please do not review the first commit.

@sbidoul : i finally didn't implemented your "maybe proposal" posted [here](https://github.com/OCA/server-env/issues/25#issuecomment-709973585). "_Maybe you could log an error if read is called (unless a specific context key is set) so we at least have a way to spot such cases ?_"
The reason is that my PR doesn't introduce bugs. In fact, if you : 
- restore a production database in a test environment
- call read method directly (without calling before ``get_param()`` function, you'll have the wrong value (production value) with the current code

Note : Do not review for the time being, tests are failing.


legalsylvain:
Now green. some simplification.
Note : I think that ``server_environment_ir_config_parameter`` should not depends on ``server_environment`` as it is working without and it has sense. but another story.

PR #69, opened 2021-01-14
[13.0] Remove required from tech_name mixin

guewen:
In some cases, we want to use a tech_name, but we do not use server
environment on all the records, so tech_name should not be required.

### DIFF
--- a/server_environment/models/server_env_tech_name_mixin.py
+++ b/server_environment/models/server_env_tech_name_mixin.py
@@ -34,8 +34,7 @@ class ServerEnvTechNameMixin(models.AbstractModel):
     # TODO: could leverage the new option for computable / writable fields
     # and get rid of some onchange / read / write code.
     tech_name = fields.Char(
-        required=True,
-        help="Unique name for technical purposes. " "Eg: server env keys.",
+        help="Unique name for technical purposes. Eg: server env keys.",
     )
 
     _server_env_section_name_field = "tech_name"


guewen:
ping @simahawk

simahawk (commented):
makes sense. This also solves the issue of the missing SQL constrain in some complex inheritance. What about the constrain at line 31? That would be broken if you don't specify it on more than 1 rec. Maybe add a functional constrain and validate only if the field is valued (which allows overrides and customization too).

guewen:
The unique constraint considers NULL as distinct values, you can have several NULL values in records:

```
odoo@localhost:testdb> create table test (id integer, name varchar);
CREATE TABLE
Time: 0.011s
odoo@localhost:testdb> create unique index test_unique on test(name);
CREATE INDEX
Time: 0.007s
odoo@localhost:testdb> insert into test(id, name) values (1, 'foo');
INSERT 0 1
Time: 0.005s
odoo@localhost:testdb> insert into test(id, name) values (2, 'foo');
duplicate key value violates unique constraint "test_unique"
DETAIL:  Key (name)=(foo) already exists.

Time: 0.005s
odoo@localhost:testdb> insert into test(id, name) values (3, NULL);
INSERT 0 1
Time: 0.003s
odoo@localhost:testdb> insert into test(id, name) values (4, NULL);
INSERT 0 1
Time: 0.003s
```

simahawk (approved):

simahawk (commented):
oh we'll need an upgrade step to remove the constrain I think

guewen:
> oh we'll need an upgrade step to remove the constrain I think

We don't change the constraint, only required :) so an upgrade is enough (I tested to be sure).

simahawk:
> I tested to be sure)

great, tnx :)

PR #75, opened 2021-07-05
[13.0] Improvements on server.env.techname.mixin

ivantodorovich:
ping @simahawk 

What do you think about this?

Re. the mixin inherit (https://github.com/OCA/server-env/commit/154090abb7a694948c2ff9a9b6809abdd85a8c0b), I guess it wasn't done like this before because one may use `server.env.techname.mixin` without `server.env.mixing` but I can't see why one would want to do that.. I find this way easier to use and in most cases saves a few line breaks

### DIFF
--- a/server_environment/models/server_env_tech_name_mixin.py
+++ b/server_environment/models/server_env_tech_name_mixin.py
@@ -2,7 +2,8 @@
 # @author Simone Orsi <simahawk@gmail.com>
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 
-from odoo import api, fields, models
+from odoo import _, api, fields, models
+from odoo.exceptions import ValidationError
 
 from odoo.addons.http_routing.models.ir_http import slugify
 
@@ -25,41 +26,58 @@ class ServerEnvTechNameMixin(models.AbstractModel):
     _sql_constraints = [
         ("tech_name_uniq", "unique(tech_name)", "`tech_name` must be unique!",)
     ]
-    # TODO: could leverage the new option for computable / writable fields
-    # and get rid of some onchange / read / write code.
+
     tech_name = fields.Char(
-        help="Unique name for technical purposes. Eg: server env keys.",
+        string="Environment Technical Name",
+        help="Unique name for server environment configuration keys.",
+        compute="_compute_tech_name",
+        store=True,
+        readonly=False,
     )
 
     _server_env_section_name_field = "tech_name"
 
-    @api.onchange("name")
-    def _onchange_name_for_tech(self):
-        # Keep this specific name for the method to avoid possible overrides
-        # of existing `_onchange_name` methods
-        if self.name and not self.tech_name:
-            self.tech_name = self.name
+    @api.depends("name")
+    def _compute_tech_name(self):
+        for rec in self:
+            # Update tech_name only if it hasn't been set or if we're
+            # dealing with a new record.
+            if not rec.tech_name or not rec._origin.id:
+                rec.tech_name = self._normalize_tech_name(rec.name)
 
     @api.onchange("tech_name")
     def _onchange_tech_name(self):
-        if self.tech_name:
-            # make sure is normalized
-            self.tech_name = self._normalize_tech_name(self.tech_name)
+        # make sure it's normalized
+        res = {}
+        normalized = self._normalize_tech_name(self.tech_name)
+        if self.tech_name != normalized:
+            res = {
+                "warning": {
+                    "title": _("Technical Name"),
+                    "message": _(
+                        "Environment Technical Name '%s' can't "
+                        "contain special characters."
+                    )
+                    % self.tech_name,
+                }
+            }
+            self.tech_name = normalized
+        return res
 
-    @api.model
-    def create(self, vals):
-        self._handle_tech_name(vals)
-        return super().create(vals)
-
-    def write(self, vals):
-        self._handle_tech_name(vals)
-        return super().write(vals)
-
-    def _handle_tech_name(self, vals):
-        # make sure technical names are always there
-        if not vals.get("tech_name") and vals.get("name"):
-            vals["tech_name"] = self._normalize_tech_name(vals["name"])
+    @api.constrains("tech_name")
+    def _check_tech_name(self):
+        for rec in self.filtered("tech_name"):
+            if rec.tech_name != self._normalize_tech_name(rec.tech_name):
+                raise ValidationError(
+                    _(
+                        "Environment Technical Name '%s' can't "
+                        "contain special characters."
+                    )
+                    % rec.tech_name
+                )
 
     @staticmethod
     def _normalize_tech_name(name):
+        if not name:
+            return name
         return slugify(name).replace("-", "_")
--- a/server_environment/models/server_env_tech_name_mixin.py
+++ b/server_environment/models/server_env_tech_name_mixin.py
@@ -16,17 +16,11 @@ class ServerEnvTechNameMixin(models.AbstractModel):
     This mixin helps solve the problem by providing a tech name field
     and a cleanup machinery as well as a unique constrain.
 
-    To use this mixin add it to the _inherit attr of your module like:
-
-        _inherit = [
-            "my.model",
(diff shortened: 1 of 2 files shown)


server_environment/models/server_env_tech_name_mixin.py
```diff
@@ -25,41 +26,36 @@ class ServerEnvTechNameMixin(models.AbstractModel):
...
-    def _handle_tech_name(self, vals):
-        # make sure technical names are always there
-        if not vals.get("tech_name") and vals.get("name"):
-            vals["tech_name"] = self._normalize_tech_name(vals["name"])
+    @api.depends("name")
+    def _compute_tech_name(self):
+        for rec in self:
+            # Update tech_name only if it hasn't been set or if we're
+            # dealing with a new record.
+            if not rec.tech_name or not rec._origin.id:
```
simahawk, line 45: I'd say only if is not set + inverse, see below

server_environment/models/server_env_tech_name_mixin.py
```diff
@@ -25,41 +26,36 @@ class ServerEnvTechNameMixin(models.AbstractModel):
...
+                rec.tech_name = self._normalize_tech_name(rec.name)
+
+    @api.constrains("tech_name")
+    def _check_tech_name(self):
+        for rec in self.filtered("tech_name"):
```
simahawk, line 69: I'd replace this w/ an inverse method that normalizes ALWAYS.

simahawk (changes requested):
> Re. the mixin inherit (154090a), I guess it wasn't done like this before because one may use server.env.techname.mixin without server.env.mixing but I can't see why one would want to do that.. I find this way easier to use and in most cases saves a few line breaks

yeah, it was in my mind for a while. At first I wanted to keep it isolated but it turns out you always use them in combo.

Thanks for taking care of my TODOs! :smile:

server_environment/models/server_env_tech_name_mixin.py
ivantodorovich: What I don't like about normalizing automatically on an inverse method is that if you write to the record programatically you'll get a silent automatic formatting, which is misleading.

ie:

```py
rec = self.env["your.model"].create({"name": "Test 123", "tech_name": "test 123"})
rec.tech_name # test_123
```

For the same reason I like to keep the constraint

What about restoring the onchange instead?

If the field is indeed exposed to the form view and written on directly, the onchange can format it automatically. We can even display a warning message.

Check my latest commit

server_environment/models/server_env_tech_name_mixin.py
ivantodorovich: My reasoning is that on new records (before they're saved) you can type a name: "My paiment providr", and then you realize you typed it incorrectly, so you correct it: "My payment provider".

I would expect the tech_name to update acordingly ("my_payment_provider"), at least until I save it.
After save, yes, it should never change again unless it's explicit.

server_environment/models/server_env_tech_name_mixin.py
simahawk: > What I don't like about normalizing automatically on an inverse method is that if you write to the record programatically you'll get a silent automatic formatting, which is misleading.

I know, but in this case I don't want to trust the user at all. At least, that's my POV ;)
 
> For the same reason I like to keep the constraint
> 
> What about restoring the onchange instead?
> 
> If the record is indeed exposed to the form view and written on directly, the onchange can format it automatically. We can even display a warning message.
> 
> Check my latest commit

I don't like the warning. As mentioned above: we should not trust the user and the user should not spend time in understanding how we want that field. This field is technical and we want take full control on it.

IMO What is relevant for the user is that the final value is displayed. That's all the user needs to know.

With this in mind, the onchange is useless if you normalize it automatically.

My $0.02

simahawk: you can keep the constraint if you prefer ;)

ivantodorovich: > I don't like the warning. As mentioned above: we should not trust the user and the user should not spend time in understanding how we want that field. This field is technical and we want take full control on it.

It's not just a warning, the onchange is also taking care of normalizing the value
The warning is to inform the user that the value has changed, rather than doing it silently (simply to play nice and not to be misleading)

As the onchange doesn't play when writing on records programatically, this enables the constraint.
Otherwise if we use inverse, the constraint will never really play and the result can be misleading in the programatic scenario.

I think how it is right now covers both scenarios: automatic on forms, exception on orm
:white_check_mark: In both cases the user isn't trusted (constraint)
:white_check_mark: In forms the field is automatically normalized (user doesn't spend time in understanding how we want the field)
:white_check_mark:  Cherry on top: warning adverting the field has been normalized 

As for how it should be exposed in forms (which is optional in fact, it could very well be hidden), I think something like this is appropiate (only **Administration / System** in debug mode):
https://github.com/OCA/server-env/blob/1c1a0b770309b3e77ebf7665a88bb92522e817c9/payment_environment/views/payment_acquirer.xml#L8-L17

Afterall this field is only meant to be seen / changed by sys admins.

server_environment/models/server_env_tech_name_mixin.py
```diff
@@ -16,56 +17,63 @@ class ServerEnvTechNameMixin(models.AbstractModel):
...
+    _inherit = "server.env.mixin"
     _description = "Server environment technical name"
     _sql_constraints = [
         ("tech_name_uniq", "unique(tech_name)", "`tech_name` must be unique!",)
     ]
-    # TODO: could leverage the new option for computable / writable fields
-    # and get rid of some onchange / read / write code.
+
     tech_name = fields.Char(
+        string="Technical Name",
```
simahawk: this does not add any extra info. I would then rename this to "Technical env name" or "Environment technical name"

server_environment/models/server_env_tech_name_mixin.py
```diff
@@ -16,56 +17,63 @@ class ServerEnvTechNameMixin(models.AbstractModel):
...
-        return super().write(vals)
-
-    def _handle_tech_name(self, vals):
-        # make sure technical names are always there
-        if not vals.get("tech_name") and vals.get("name"):
-            vals["tech_name"] = self._normalize_tech_name(vals["name"])
+        # make sure it's normalized
+        res = {}
+        normalized = self._normalize_tech_name(self.tech_name)
+        if self.tech_name != normalized:
```
simahawk, line 53: I still feel that this useless and we can simply describe the behavior in the `help` attribute. But... yeah whatever :sweat_smile:

server_environment/models/server_env_tech_name_mixin.py
```diff
@@ -16,56 +17,63 @@ class ServerEnvTechNameMixin(models.AbstractModel):
...
+            if rec.tech_name != self._normalize_tech_name(rec.tech_name):
+                raise ValidationError(
+                    _("Technical Name '%s' can't contain special characters.")
+                    % rec.tech_name
+                )
 
     @staticmethod
     def _normalize_tech_name(name):
+        if not name:
+            return False
```
simahawk:
```suggestion
            return ""
```

simahawk:
@ivantodorovich could you test this on OCA/search-engine ?

simahawk:
@ivantodorovich ping :)

ivantodorovich:
> @ivantodorovich could you test this on OCA/search-engine ?

It seems ok
Is there something specific you have in mind?

simahawk:
just use this change there and see if you get a green build ;)

ivantodorovich:
> just use this change there and see if you get a green build ;)

Indeed it's not working.

`connecto_search_engine` would need to be adapted to have the `index_prefix_name` in a computed field.

But also I tried that, and it's not working as expected. Since `server_environment` "transforms" fields into computed fields to read from config. I created this issue to explain it: https://github.com/OCA/server-env/issues/85

I couldn't find a simple solution for now :/ so best if we close this and reevaluate in a future migration IMO

simahawk:
@ivantodorovich you can reconsider this here https://github.com/OCA/server-env/pull/122 :wink:

PR #76, opened 2021-07-05
[13.0][ADD] payment_environment

ivantodorovich:
Configure payment acquirers with server_environment

### DIFF
--- a/payment_environment/__manifest__.py
+++ b/payment_environment/__manifest__.py
@@ -1,0 +1,16 @@
+# Copyright 2021 Camptocamp SA
+# @author Iván Todorovich <ivan.todorovich@gmail.com>
+# License LGPL-3.0 or later (https://www.gnu.org/licenses/lgpl.html)
+
+{
+    "name": "Payment Acquirers' configuration with server_environment",
+    "summary": "Configure payment acquirers with server_environment",
+    "category": "Tools",
+    "version": "13.0.1.0.0",
+    "license": "LGPL-3",
+    "author": "Camptocamp, Odoo Community Association (OCA)",
+    "website": "https://github.com/OCA/server-env",
+    "maintainers": ["ivantodorovich"],
+    "depends": ["payment", "server_environment"],
+    "data": ["views/payment_acquirer.xml"],
+}
--- a/payment_environment/models/payment_acquirer.py
+++ b/payment_environment/models/payment_acquirer.py
@@ -1,0 +1,23 @@
+# Copyright 2021 Camptocamp SA
+# @author Iván Todorovich <ivan.todorovich@gmail.com>
+# License LGPL-3.0 or later (https://www.gnu.org/licenses/lgpl.html)
+
+from odoo import models
+
+
+class PaymentAcquirer(models.Model):
+    _name = "payment.acquirer"
+    _inherit = [
+        "payment.acquirer",
+        "server.env.techname.mixin",
+        "server.env.mixin",
+    ]
+
+    @property
+    def _server_env_fields(self):
+        base_fields = super()._server_env_fields
+        acquirer_fields = {
+            "state": {},
+        }
+        acquirer_fields.update(base_fields)
+        return acquirer_fields
--- a/payment_environment/views/payment_acquirer.xml
+++ b/payment_environment/views/payment_acquirer.xml
@@ -1,0 +1,18 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<!--
+    Copyright 2021 Camptocamp SA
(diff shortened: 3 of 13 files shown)


payment_environment/models/payment_acquirer.py
```diff
@@ -0,0 +1,23 @@
+# Copyright 2021 Camptocamp SA
+# @author Iván Todorovich <ivan.todorovich@gmail.com>
+# License LGPL-3.0 or later (https://www.gnu.org/licenses/lgpl.html)
```
simahawk, line 3: AGPL or LGPL? :)

simahawk: after checking it seems there's a bit of confusion on licenses. 

The main module says GPL but refers to AGPL link, the tech mixing [says AGPL](https://github.com/OCA/server-env/blob/13.0/server_environment/models/server_env_mixin.py#L2) but seems wrong to me.

Would you mind opening another PR to fix them so that we can fast track it?

LGPL seems the right one for the whole thing. Maybe ping @guewen  on the new PR.

payment_environment/models/payment_acquirer.py
```diff
@@ -0,0 +1,23 @@
...
+
+from odoo import models
+
+
+class PaymentAcquirer(models.Model):
+    _name = "payment.acquirer"
+    _inherit = [
+        "payment.acquirer",
+        "server.env.techname.mixin",
```
simahawk, line 12: you can save a line here tnx to https://github.com/OCA/server-env/pull/75

simahawk (changes requested):

payment_environment/models/payment_acquirer.py
ivantodorovich: PR created here: https://github.com/OCA/server-env/pull/77

As for this, my intention was AGPL, this got carried on from a copy & paste left over.
Fixing it now :sweat_smile:

payment_environment/models/payment_acquirer.py
ivantodorovich: Indeed :) but I'll wait until it's merged

simahawk:
#77 is getting merged

ivantodorovich:
Thanks! PR updated accordignly :)

payment_environment/__manifest__.py
```diff
@@ -0,0 +1,16 @@
+# Copyright 2021 Camptocamp SA
+# @author Iván Todorovich <ivan.todorovich@gmail.com>
+# License LGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html)
```
simahawk:
```suggestion
# License LGPL-3.0 or later (https://www.gnu.org/licenses/lgpl.html)
```
? :smile:

payment_environment/models/payment_acquirer.py
```diff
@@ -0,0 +1,23 @@
+# Copyright 2021 Camptocamp SA
+# @author Iván Todorovich <ivan.todorovich@gmail.com>
+# License LGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html)
```
simahawk: same

payment_environment/views/payment_acquirer.xml
```diff
@@ -0,0 +1,18 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<!--
+    Copyright 2021 Camptocamp SA
+    @author Iván Todorovich <ivan.todorovich@gmail.com>
+    License LGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html)
```
simahawk: guess what? :D

ivantodorovich: :(

ivantodorovich: done!

simahawk (commented):
@ivantodorovich I've just realized that this should probably go to https://github.com/OCA/?q=payment

simahawk:
@ivantodorovich ping :)

ivantodorovich:
>  @ivantodorovich I've just realized that this should probably go to https://github.com/OCA/?q=payment

Should it thought? 🤔 

by the same logic, shouldn't `mail_environment` go to https://github.com/OCA/social; `pos_environment` go to https://github.com/OCA/pos; `carrier_environment` go to https://github.com/OCA/delivery-carrier; etc..

This is kind of a gray area I think 😅

gurneyalex:
I would keep this in OCA/server-env if you start using it, it's nice to have all the deps in 1 single place.

gurneyalex (approved):

PR #77, opened 2021-07-06
[UPD] Change LICENSE to LGPL

ivantodorovich:
ping @simahawk @guewen 

Following up on your comment here: https://github.com/OCA/server-env/pull/76#discussion_r664276443

> after checking it seems there's a bit of confusion on licenses.
>
>The main module says GPL but refers to AGPL link, the tech mixing says AGPL but seems wrong to me.
>
>Would you mind opening another PR to fix them so that we can fast track it?

Would you mind point out what you saw?
If it's indeed LGPL what we need here, do we also need to change the license of the modules that depend on it?

Also what about this file?
If it's based on someone else's work I don't think we can change it, right?
https://github.com/OCA/server-env/blob/314d16d37c8ce56603c3d1ad9c2ff596d1f22425/server_environment/server_env.py#L1-L19

### DIFF
--- a/server_environment/__init__.py
+++ b/server_environment/__init__.py
@@ -1,22 +1,3 @@
-##############################################################################
-#
-#    Adapted by Nicolas Bessi. Copyright Camptocamp SA
-#    Based on Florent Xicluna original code. Copyright Wingo SA
-#
-#    This program is free software: you can redistribute it and/or modify
-#    it under the terms of the GNU General Public License as published by
-#    the Free Software Foundation, either version 3 of the License, or
-#    (at your option) any later version.
-#
-#    This program is distributed in the hope that it will be useful,
-#    but WITHOUT ANY WARRANTY; without even the implied warranty of
-#    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
-#    GNU General Public License for more details.
-#
-#    You should have received a copy of the GNU General Public License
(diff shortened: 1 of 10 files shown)


simahawk (changes requested):
regarding `GPL` change we have to ask @nbessi and @guewen.

My remark  was that the header was referring to `GPL` but linking the `AGPL` license.

Very likely there was a reason for going w/ GPL. For the tech mixin I don't care that much but it should probably be the same as the rest of the module.

CC @gurneyalex

guewen:
I have no idea why originally it was GPL, but AFAIR this module dates from OpenERP 5.0, which was GPL.
Considering 

>    Adapted by Nicolas Bessi. Copyright Camptocamp SA 
>    Based on Florent Xicluna original code. Copyright Wingo SA 

I suppose @nbessi and @florentx (hi!) may know better than me if we can change the license :)
Apparently, the code that goes back from the original module is mainly the one that generates the view dynamically https://github.com/OCA/server-env/blob/314d16d37c8ce56603c3d1ad9c2ff596d1f22425/server_environment/server_env.py#L176-L315 and system info https://github.com/OCA/server-env/blob/314d16d37c8ce56603c3d1ad9c2ff596d1f22425/server_environment/system_info.py

nbessi:
Purely for historical reason. Fine for me we can move to LGPL.

sbidoul (approved):

simahawk (approved):

PR #80, opened 2021-10-05
[EXP] experimental try to run 14 version on 15 version

sebastienbeau:
Experimental test to see if the current version work on v15 version without changes.


sebastienbeau:
conclusion version 14 is not compatible with version 15, as there is some change in orm that break the dynamic build of "server.config"

PR #83, opened 2021-10-18
[MIG] server_environment: migration to 15.0

JuMiSanAr:
Migrate server_environment to version 15.

PR #87, opened 2021-10-22
 [MIG] server_environment: migration to 15.0

simahawk:
Replaces #83 (already approved) to work around a pre-commit cache issue on github actions.


sbidoul:
@simahawk Can you check if the generated .pot file looks good ? It seems it [removed](https://github.com/OCA/server-env/commit/a883ca205cb340f1047f59b6ce5982c52dcffd62) a lot of stuff.

simahawk:
Those strings seem useless (to me at least) as they are all technical.

PR #88, opened 2021-10-22
[REF] .t2d.yml: Fix runbot red

moylop260:
Docker image used from runbot is too old
So it is using python3.7 as last python version installed
But only 4.5.4 coverage version is supported by this version of py

### DIFF
--- a/.t2d.yml
+++ b/.t2d.yml
@@ -5,10 +5,10 @@ cache:
     - $HOME/.cache/pre-commit
 
 python:
-  - "3.8"
+  - "3.7"
 
 addons:
-  postgresql: "9.6"
+  postgresql: "11"
   apt:
     packages:
       - expect-dev # provides unbuffer utility
@@ -27,13 +27,13 @@ jobs:
 env:
   global:
     - VERSION="15.0" TESTS="0" LINT_CHECK="0" MAKEPOT="0"
-    - MQT_DEP=PIP
 
 install:
   - git clone --depth=1 https://github.com/OCA/maintainer-quality-tools.git
     ${HOME}/maintainer-quality-tools
   - export PATH=${HOME}/maintainer-quality-tools/travis:${PATH}
   - travis_install_nightly
+  - pip install -U coverage==4.5.4
 
 script:
   - travis_run_tests


pedrobaeza:
But this problem is going to be present on all 15.0 branches, as we want to use Python 3.8 on them. Can we bump the travis2docker version instead?

gurneyalex:
> But this problem is going to be present on all 15.0 branches, as we want to use Python 3.8 on them. Can we bump the travis2docker version instead?

FWIU the issue is more in the base docker image we use to run things on, which does not have python3.8 available.

I think it would be a very good idea to upgrade it, esp. since it uses a very old Ubuntu as base version (see https://github.com/Vauxoo/docker-ubuntu-base/blob/master/Dockerfile)

pedrobaeza:
Or maybe we can use new runbot docker system directly.

moylop260:
Both ideas are good

This PR is only a quick workaround in order to have the runbot v15.0 branches green ASAP

I'm not able to work on this right now

But feel free to take this job to use a newer global image or using runbot based on docker images

It could be great!

moylop260:
BTW runbot is green now for this project:
 - Screen Shot 2021-10-22 at 11 27 53


cc @sbidoul @Yajo

gurneyalex:
@sbidoul do you think we could relax the python version requirements on the pypi published packages to >= 3.7 ?

sbidoul:
Now that Odoo [finally says](https://github.com/odoo/odoo/blob/1fad826743de9c8f916ef083e8de453212df6959/setup.py#L59) >= 3.7, I think we can, but we should then also test on python 3.7 on the OCA CI.

sbidoul:
The discussion that led to chosing 3.8 is in https://github.com/OCA/oca-addons-repo-template/pull/70 (in the resolved comments)

gurneyalex:
> Now that Odoo [finally says](https://github.com/odoo/odoo/blob/1fad826743de9c8f916ef083e8de453212df6959/setup.py#L59) >= 3.7, I think we can, but we should then also test on python 3.7 on the OCA CI.

we would, at least with runbot, have 3.7. So you mean changing the github actions setup to use 3.7 too?

sbidoul:
I created https://github.com/OCA/oca-addons-repo-template/pull/84 and friends to test on 3.7.

PR #82, opened 2021-10-18
[MIG] 15.0 mail_environment

JuMiSanAr:
Migrate mail_environment to version 15.

PR #89, opened 2021-10-27
[FIX] Update server_env_mixin.py to allow mail_environment migration

JuMiSanAr:
Fixup to server_environment to allow prevent error NoneType in server_env_mixin.py.
This change is necessary to proceed with PR [#82](https://github.com/OCA/server-env/pull/82).

Context:

fields.arg can be None thus `copy()` will fail.
Those circumstances are met when adding module `mail_environment`

### DIFF
--- a/server_environment/models/server_env_mixin.py
+++ b/server_environment/models/server_env_mixin.py
@@ -376,7 +376,7 @@ class ServerEnvMixin(models.AbstractModel):
         # (inherits), we want to override it with a new one
         if fieldname not in self._fields or self._fields[fieldname].inherited:
             base_field_cls = base_field.__class__
-            field_args = base_field.args.copy()
+            field_args = base_field.args.copy() if base_field.args else {}
             field_args.pop("_sequence", None)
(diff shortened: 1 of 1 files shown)


server_environment/models/server_env_mixin.py
```diff
@@ -390,7 +390,10 @@ def _setup_base(self):
         super()._setup_base()
         for fieldname in self._server_env_fields:
             field = self._fields[fieldname]
-            self._server_env_add_default_field(field)
+            if not field.args:
```
simahawk: any comment on the motivation?

JuMiSanAr: The checks for PR #82 failed because of the below:

JuMiSanAr: https://github.com/OCA/server-env/runs/4019208495?check_suite_focus=true

yvaucher (approved):

simahawk (approved):

PR #93, opened 2021-12-02
14.0 server_environment is Production/Stable

gurneyalex:
server_environment is a dependency of other modules which are in this status
and now that we are checking this, we need to declare the maturity
explicitly.

### DIFF
--- a/server_environment/__manifest__.py
+++ b/server_environment/__manifest__.py
@@ -10,6 +10,7 @@
     "summary": "move some configurations out of the database",
     "website": "https://github.com/OCA/server-env",
     "license": "LGPL-3",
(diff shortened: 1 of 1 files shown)


simahawk (approved):
rebase needed

yvaucher (approved):

simahawk (changes requested):
pls cherry-pick the commit from 15 which is already merged, to avoid false positives on fwd/back porting

gurneyalex:
I also cherry picked the licence change commit  that was not here

gurneyalex:
not sure why runboat is red, but I think we can merge anyway

PR #95, opened 2021-12-02
13.0 server_environment is Production/Stable

gurneyalex:
server_environment is a dependency of other modules which are in this status
and now that we are checking this, we need to declare the maturity
explicitly


yvaucher (approved):

simahawk (changes requested):
pls cherry-pick the commit from 15 which is already merged, to avoid false positives on fwd/backport

PR #96, opened 2021-12-15
[14.0][FIX] payment_environment: Fix state usage

jcoux:
As state field is now managed as server environment fields, the field is considered as a computed fields.
Then, we need to define a custom search function to be able to search on this field.

### DIFF
--- a/payment_environment/models/payment_acquirer.py
+++ b/payment_environment/models/payment_acquirer.py
@@ -2,7 +2,8 @@
 # @author Iván Todorovich <ivan.todorovich@gmail.com>
 # License LGPL-3.0 or later (https://www.gnu.org/licenses/lgpl.html)
 
-from odoo import models
+from odoo import fields, models
+from odoo.osv import expression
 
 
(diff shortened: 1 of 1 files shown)


payment_environment/models/payment_acquirer.py
```diff
@@ -21,3 +22,33 @@ def _server_env_fields(self):
...
+        See module payment in controller/portal.py function pay()
+
+        Used domain is: ('state', 'in', ['enabled', 'test'])
+        """
+        if operator == "in" and isinstance(value, list):
+            valid_ids = []
+            all_acquirers = self.with_context(active_test=False).search([])
+            for acquirer in all_acquirers:
+                if acquirer.state in value:
+                    valid_ids.append(acquirer.id)
```
gurneyalex, line 50:
```suggestion
            valid_ids = all_acquirers.filtered_domain([('state', 'in', value)])
```

gurneyalex (approved):

PR #97, opened 2021-12-20
[14.0][ADD] server_environment_iap: to override IAP Accounts from server environment file

yankinmax:
It would be nice to set IAP Accounts in the server environment file, possibly make their key and value
readonly in the user interface and remove them from the database except for production.

### DIFF
--- a/server_environment_iap/models/iap_account.py
+++ b/server_environment_iap/models/iap_account.py
@@ -1,70 +1,31 @@
 # Copyright 2021 Camptocamp SA <https://www.camptocamp.com/>
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl)
 
-from odoo import _, api, fields, models
-from odoo.exceptions import UserError
-
-from odoo.addons.server_environment.server_env import serv_config
-
-SECTION = "iap.account"
+from odoo import api, models
 
 
 class IapAccount(models.Model):
+    _name = "iap.account"
+    _inherit = [
+        "iap.account",
+        "server.env.techname.mixin",
+        "server.env.mixin",
+    ]
 
-    _inherit = "iap.account"
-
-    is_environment = fields.Boolean(
-        string="Defined by environment",
-        compute="_compute_is_environment",
-        help="If check, the value in the database will be ignored"
-        " and alternatively, the system will use the service name defined"
-        " in your odoo.cfg environment file.",
-    )
-
-    def _compute_is_environment(self):
-        for account in self:
-            account.is_environment = serv_config.has_option(
-                SECTION, account.service_name
-            )
-
-    @api.model
-    def get(self, service_name, force_create=True):
-        account = super().get(service_name, force_create=True)
-        if serv_config.has_option(SECTION, service_name):
-            cvalue = serv_config.get(SECTION, service_name)
-            if not cvalue:
-                # if service name is empty it's probably not a production instance,
-                # so we need to remove it from database
-                account.unlink()
-                raise UserError(
-                    _("Service name %s is empty in " "server_environment_file")
-                    % (service_name,)
-                )
-            if cvalue != account.account_token:
-                # we write in db on first access;
-                # should we have preloaded values in database at,
-                # server startup, modules loading their parameters
-                # from data files would break on unique key error.
-                account.account_token = cvalue
-        return account
+    @property
+    def _server_env_fields(self):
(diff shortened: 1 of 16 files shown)


yankinmax:
@gurneyalex take a look pls here. Is my workflow correct?

gurneyalex:
@yankinmax any reason not to use the server.environment.mixin? (see https://github.com/OCA/server-env/blob/14.0/server_environment/models/server_env_mixin.py for the documentation)

server_environment_iap/readme/CONFIGURE.rst
```diff
@@ -0,0 +1,14 @@
+With this module installed, the IAP Accounts are configured
+in the `server_environment_files` module (which is a module
```
simahawk: this is not mandatory, you can use `SERVER_ENV_CONFIG` + `SERVER_ENV_CONFIG_SECRET` via env vars.
I'd simply link to `server_environment` README.

server_environment_iap/readme/CREDITS.rst
```diff
@@ -0,0 +1,2 @@
+This module is maintained by:
```
simahawk: remove this file, is not needed

server_environment_iap/readme/USAGE.rst
```diff
@@ -0,0 +1,3 @@
+Once configured, Odoo will read the IAP Accounts values from the
```
simahawk: not needed IMO

server_environment_iap/views/iap_views.xml
```diff
@@ -0,0 +1,15 @@
...
+<odoo>
+    <record id="iap_account_view_form" model="ir.ui.view">
+        <field name="name">iap.account.form.inherit</field>
+        <field name="model">iap.account</field>
+        <field name="inherit_id" ref="iap.iap_account_view_form" />
+        <field name="groups_id" eval="[(4, ref('base.group_system'))]" />
+        <field name="arch" type="xml">
+            <field name="service_name" position="before">
+                <field name="name" groups="base.group_no_one" />
+                <!-- <field name="tech_name" groups="base.group_no_one" /> -->
```
simahawk: tech_name must be visible

server_environment_iap/models/iap_account.py
```diff
@@ -0,0 +1,33 @@
...
+
+class IapAccount(models.Model):
+    _name = "iap.account"
+    _inherit = [
+        "iap.account",
+        "server.env.techname.mixin",
+        "server.env.mixin",
+    ]
+
+    name = fields.Char("Name", required=True)
```
simahawk: this field is not needed. The original model has `service_name` and the `tech_name` is the one used for env conf.

simahawk (changes requested):

server_environment_iap/models/iap_account.py
yankinmax: probably I don't understand something, but without this field it doesn't work as I expect)
So, I need `conf` file to include such iap.account configuration:
```
[iap_account.resource_name]
service_name = my_service_name
account_token = my_secret_token
```
How should I use `tech_name` field then?
This code works like this: `name` becomes `tech_name` and unique, so then if I create new iap.account in UI `service_name` and `account_token` are filled with values from `conf`.
I can't use `service_name` as `tech_name` as due to source code iap accounts are company dependent, so we can have for example: partner_autocomplete in Company A and partner_autocomplete in Company B. Both service names are correct.
If you can point me where I'm wrong I would highly appreciate it. Thanks

yankinmax: I got it. Sorry for the noise.

server_environment_iap/tests/test_server_environment_iap.py
```diff
@@ -0,0 +1,87 @@
...
+                {
+                    "tech_name": "account_4",
+                    "service_name": "new_partner_autocomplete",
+                    "account_token": "my_new_secret_token",
+                }
+            )
+            self.assertEqual(account.service_name, "new_partner_autocomplete")
+            self.assertEqual(account.account_token, "my_new_secret_token")
+
+    # TODO: should it be overriden on xml import?
```
yankinmax, line 64: @gurneyalex need your help to finish with tests, can you pls take a look?

gurneyalex (approved):

PR #98, opened 2021-12-22
[FIX] server_environment_ir_config_parameter

gurneyalex:
fix a conflict between this module and mail which overrides the reading
of some mail related system parameters

### DIFF
--- a/server_environment_ir_config_parameter/models/ir_config_parameter.py
+++ b/server_environment_ir_config_parameter/models/ir_config_parameter.py
@@ -39,8 +39,18 @@ class IrConfigParameter(models.Model):
                 # should we have preloaded values in database at,
                 # server startup, modules loading their parameters
                 # from data files would break on unique key error.
-                self.sudo().set_param(key, cvalue)
-                value = cvalue
+                if not self.env.context.get("_from_get_param", 0):
+                    # the check is to avoid recursion, for instance the mail
+                    # addon has an override in ir.config_parameter::write which
+                    # calls get_param if we are setting mail.catchall.alias and
+                    # this will cause an infinite recursion. We cut that
+                    # recursion by using the context check.
+                    #
+                    # The mail addon call to get_param expects to get the value
+                    # *before* the change, so we have to return the database
+                    # value in that case
+                    self.sudo().with_context(_from_get_param=1).set_param(key, cvalue)
+                    value = cvalue
         if value is None:
             return default
         return value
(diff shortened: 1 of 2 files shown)


server_environment_ir_config_parameter/models/ir_config_parameter.py
```diff
@@ -39,8 +39,18 @@ def get_param(self, key, default=False):
...
+                if not self.env.context.get("_from_get_param", 0):
+                    # the check is to avoid recursion, for instance the mail
+                    # addon has an override in ir.config_parameter::write which
+                    # calls get_param if we are setting mail.catchall.alias and
+                    # this will cause an infinite recursion. We cut that
+                    # recursion by using the context check.
+                    #
+                    # The mail addon call to get_param expects to get the value
+                    # *before* the change, so we have to return the database
+                    # value in that case
```
grindtildeath, line 51: I'm just not sure about that one, because the call to `get_param` overwrites the value that is used in the `write` method: https://github.com/odoo/odoo/blob/14.0/addons/mail/models/ir_config_parameter.py#L20

In such case, AFAIU the value in DB will always be used instead of the one defined by server environment, which will never be used since it will happen anytime `get_param` is called for `mail.catchall.alias`.

I feel like we should return `cvalue` if the context key is set...

gurneyalex: I tried this, but it does not work. If you do that, you get an error saying that the alias is already used. See https://github.com/odoo/odoo/blob/0e723c6c15dc8cd178625cf726f038ebd14e5c75/addons/mail/models/mail_alias.py#L153-L154

grindtildeath: So if I understand correctly the `get_param` on `mail.catchall.alias` will never provide the value in server environment? Shouldn't we instead add a glue module with `mail` to ensure it will read this value instead of the one in database?

grindtildeath: @gurneyalex If the goal of this piece of code is to set the value in DB, but using ORM `write` creates a recursion, shouldn't we execute an `UPDATE` using SQL?

santostelmo (approved):

server_environment_ir_config_parameter/models/ir_config_parameter.py
gurneyalex: @grindtildeath using SQL: maybe... OTOH, the PR I have here works, so...

sbidoul (approved):

ivantodorovich (approved):

PR #103, opened 2022-03-18
server_environment: Update label of env default field to avoid warnin…

adrienpeiffer:
…g messages

### DIFF
--- a/server_environment/models/server_env_mixin.py
+++ b/server_environment/models/server_env_mixin.py
@@ -378,7 +378,14 @@ class ServerEnvMixin(models.AbstractModel):
             base_field_cls = base_field.__class__
             field_args = base_field.args.copy() if base_field.args else {}
             field_args.pop("_sequence", None)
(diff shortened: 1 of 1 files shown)


sbidoul (approved):
The warning we want to silence is `odoo.addons.base.models.ir_model: Two fields (smtp_host_env_default, smtp_host) of ir.mail_server() have the same label: SMTP Server`

sbidoul:
@simahawk @gurneyalex

lmignon (approved):

simahawk (approved):
LG. @adrienpeiffer pls next time keep the commit msg short and move the long description to the description of the commit.

See also https://www.odoo.com/documentation/15.0/developer/misc/other/guidelines.html#commit-message-structure

simahawk:
I guess this should be backported to older versions?
If yes, pls rewrite the commit msg 1st otherwise when we fwd/backport commits we'll always have a false positive.

adrienpeiffer:
@simahawk Commit updated. Thanks

PR #104, opened 2022-03-25
[Fix] server env tech name unique constraint

santostelmo:
Default `tech_name` field is filled from record `name`.
We have an issue with multi companies. For example, if we tr to set the char of accounts for 2 companies, jornal names and tech_names are populated with the same values in both companies. To avoid that for models defined per company, we change the `tech_name` unique constraint rule.

### DIFF
--- a/server_environment/models/server_env_tech_name_mixin.py
+++ b/server_environment/models/server_env_tech_name_mixin.py
@@ -2,7 +2,8 @@
 # @author Simone Orsi <simahawk@gmail.com>
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 
-from odoo import api, fields, models
+from odoo import api, fields, models, _
+from odoo.exceptions import ValidationError
 
 from odoo.addons.http_routing.models.ir_http import slugify
 
@@ -28,9 +29,6 @@ class ServerEnvTechNameMixin(models.AbstractModel):
 
     _name = "server.env.techname.mixin"
     _description = "Server environment technical name"
-    _sql_constraints = [
-        ("tech_name_uniq", "unique(tech_name)", "`tech_name` must be unique!",)
-    ]
     # TODO: could leverage the new option for computable / writable fields
     # and get rid of some onchange / read / write code.
     tech_name = fields.Char(
@@ -38,7 +36,20 @@ class ServerEnvTechNameMixin(models.AbstractModel):
     )
 
     _server_env_section_name_field = "tech_name"
+    _tech_name_unique_per_company = False   # off by default
 
+    @api.constrains("tech_name", "company_id")
+    def _check_tech_name(self):
+        company_dependent = self._tech_name_unique_per_company and  "company_id" in self._fields
+        for record in self:
+            domain = [('id', '!=', record.id), ('tech_name', '=' , record.tech_name)]
+            error_msg = _("Tech name %s is duplicated. Tech name must be unique") % record.tech_name
+            if company_dependent:
+                domain.append(('company_id', '=', record.company_id.id))
+                error_msg += _(" per company")
+            if self.search_count(domain):
+                raise ValidationError(error_msg)
+                
     @api.onchange("name")
     def _onchange_name_for_tech(self):
         # Keep this specific name for the method to avoid possible overrides


server_environment/models/server_env_tech_name_mixin.py
```diff
@@ -28,9 +29,6 @@ class ServerEnvTechNameMixin(models.AbstractModel):
 
     _name = "server.env.techname.mixin"
     _description = "Server environment technical name"
-    _sql_constraints = [
```
simahawk, line 31: I disagree w/ this change. 
The main reason being: the name must be unique per ENV not per company -> how do you make sure you set the right name for different companies in the same env?
You must make sure that were you use this mixin the tech name is computed properly (eg: by prepending the company name or a tech name for the company).

simahawk (changes requested):

server_environment/models/server_env_tech_name_mixin.py
santostelmo: I considered to prepend the company name in case of duplication, but I didn't liked the option of letting the method _normalize_tech_name appliy the transformation  `tech_name = company_name + "_" + record.name`
But it is still an option. :+1: 

Why a tech_name can not be the same if the record belongs to different companies ? Question: If I have an environment value, that value can not be applied to multiple records?

simahawk: IMO it should be explicit and you should have 2 entries in the env conf.

Regarding the normalization of the name: it's better to have a tech field on the company too because you don't want to lose the conf if someone changes the name (that's why "tech_name" is used here). You could use the ID of the company if you prefer.

server_environment/models/server_env_tech_name_mixin.py
```diff
@@ -39,6 +37,15 @@ class ServerEnvTechNameMixin(models.AbstractModel):
...
 
+    @api.constrains('tech_name')
+    def _check_tech_name(self):
+
+        for record in self:
+            if hasattr(record, 'company_id') and self.search([('id', '!=', record.id), ('tech_name', '=' , record.tech_name), ('company_id', '=', record.company_id.id)]):
+                raise ValidationError(_("Tech name %s is duplicated. Tech name must be unique per company.") % record.tech_name)
+            if not hasattr(record, 'company_id') and self.search([('id', '!=', record.id), ('tech_name', '=' , record.tech_name)]):
+                raise ValidationError(_("Tech name %s is duplicated. Tech name must be unique.") % record.tech_name)
+
```
simahawk: If you want to keep your proposal please adjust it to be similar to this

```suggestion
    _tech_name_unique_per_company = False   # off by default
    [...]
    @api.constrains("tech_name", "company_id")
    def _check_tech_name(self):
        company_dependent = self._tech_name_unique_per_company and  "company_id" in self._fields
        for record in self:
            domain = [('id', '!=', record.id), ('tech_name', '=' , record.tech_name)]
            error_msg = _("Tech name %s is duplicated. Tech name must be unique") % record.tech_name
            if company_dependent:
                domain.append(('company_id', '=', record.company_id.id))
                error_msg += _(" per company")
            if self.search_count(domain):
                raise ValidationError(error_msg)
```
and then add a test :pray:

simahawk (changes requested):

server_environment/models/server_env_tech_name_mixin.py
santostelmo: Good idea :+1:  I applied that code

simahawk: I'm afraid this piece of code is not enough. Existing installation will have the constraint in the DB.
We need an upgrade step that looks into all inheriting models and drops the constraint.
Furthermore, I'd like to see some tests.
If this is too much for you/your project right now, I would add a TODO w/ a reference to this PR and stick to a proper computation of the name using the company ID or a special field.

santostelmo: I dropped the constraint in the project where I am using that code. I'll spend some time to apply this improvements

santostelmo:
@simahawk keeping the original code and doing that could be acceptable for you ?
```
    def _handle_tech_name(self, vals):
        # make sure technical names are always there
        if not vals.get("tech_name") and vals.get("name"):
            name = vals.get("name")
            if self.search_count([('name', '=', name)]) >= 1 and vals.get("company_id"):
                # Name suffix with company ID to avoid duplicated technames in multy company
                name += "-" +  str(vals["company_id"])
            vals["tech_name"] = self._normalize_tech_name(name)
```

PR #105, opened 2022-04-17
Update README.rst

klodr:
typo


rvalyi (approved):

rvalyi:
Trivial, merging...
/ocabot merge nobump

rvalyi:
my bad, I checked too fast. @klodr thanks for the fix attempt. However this very README.rst file is generated automatically when we merge something in the module. So you shouldn't touch this file but instead you should alter the `__manifest__.py` which is the place where is the typo https://github.com/klodr/server-env/blob/patch-1/auth_saml_environment/__manifest__.py#L5
could you do it and amend your commit please?

gurneyalex:
@rvalyi please don't use nobump unless the PR already updated the version: this causes issues with the upload to pypi (different package with same version). merge patch is what we need in this case.

rvalyi:
> @rvalyi please don't use nobump unless the PR already updated the version: this causes issues with the upload to pypi (different package with same version). merge patch is what we need in this case. 

Interesting. I had no idea about this. I see PSCs using nobump all over the place in many repos... I doubt people are aware about this pypi issue. Maybe we should change the nobump keyword to catch the PSCs attention about this and force them to think twice?

In this precise case however the PR couldn't be merged because it was trying to patch the wrong .rst file as I said in my previous comment.

simahawk:
@rvalyi @gurneyalex let's clarify: you could use nobump. The version on pypi will be `*devXYZ`.

However, you are still doing a small change - no matter how little it is - thus is fine and somehow desired to use `patch` instead.

simahawk:
here the build fails w/ 
```
odoo.exceptions.UserError: Unable to install module "auth_saml" because an external dependency is not met: Unable to find 'xmlsec1' in path
2022-04-19 06:41:49,087 23495 INFO openerp_template odoo.service.server: Initiating shutdown 
2022-04-19 06:41:49,087 23495 INFO openerp_template odoo.service.server: Hit CTRL-C again or send a second signal to force the shutdown. 
```

rafaelbn:
@rousseldenis comment me something around `pypi` it here https://github.com/OCA/sale-workflow/pull/1897#issuecomment-1048128195 and here https://github.com/OCA/sale-workflow/pull/1897#issuecomment-1048631227

cc @yajo @Shide

rousseldenis:
@gurneyalex Any advice on this ? What is missing for xmlsec1 package ?

ivantodorovich (approved):

rousseldenis:
> @rvalyi @gurneyalex let's clarify: you could use nobump. The version on pypi will be `*devXYZ`.
> 
> However, you are still doing a small change - no matter how little it is - thus is fine and somehow desired to use `patch` instead.

Yes but will be marked as pre-release

ivantodorovich:
CI should be green if you rebase now

PR #106, opened 2022-04-26
[FIX] payment_environment: Fix state usage

santostelmo:
As state field is now managed as server environment fields, the field is considered as a computed fields.
Then, we need to define a custom search function to be able to search on this field.

back port https://github.com/OCA/server-env/pull/96


gurneyalex (approved):
pre approving. 

Can you also change the _order attribute of the model to `''module_state, sequence, name'` (i.e. remove `state` from the list)?

PR #108, opened 2022-04-26
[14.0] [FW] payment_environment: Remove state from payment acquirer order

ivantodorovich:
forward port https://github.com/OCA/server-env/commit/2f17a86de140cc59884e9eaef58ac3549253cca4


ivantodorovich:
Travis and runboat are both failing but not related to this pr

simahawk (approved):

simahawk:
still `odoo.exceptions.UserError: Unable to install module "auth_saml" because an external dependency is not met: Unable to find 'xmlsec1' in path`

Issue #110, opened 2022-04-27
Runboat broken server_environment_data_encryption


simahawk:
This PR seems to solve one issue, but there's another one in runboat

`odoo.exceptions.ValidationError: No 'encryption_key_test' entry found in config file. Use a key similar to: b'lUMc38soCfqymcvsJdsXYAB_DQGZozQQYPYHOZ2PZik='`

Seems related to `server_environment_data_encryption` module. Do you have any idea what could be wrong?
@florian-dacosta @sebastienbeau 

<details>
<summary>Full traceback</summary>

```
2022-04-27 11:56:32,531 77 INFO b7b3579b4-0fcf-4ca0-8b98-4b272b74d91e odoo.modules.loading: loading payment_transfer/data/payment_acquirer_data.xml 
2022-04-27 11:56:32,541 77 WARNING b7b3579b4-0fcf-4ca0-8b98-4b272b74d91e odoo.modules.loading: Transient module states were reset 
2022-04-27 11:56:32,541 77 ERROR b7b3579b4-0fcf-4ca0-8b98-4b272b74d91e odoo.modules.registry: Failed to load registry 
2022-04-27 11:56:32,541 77 CRITICAL b7b3579b4-0fcf-4ca0-8b98-4b272b74d91e odoo.service.server: Failed to initialize database `b7b3579b4-0fcf-4ca0-8b98-4b272b74d91e`. 
Traceback (most recent call last):
  File "/opt/odoo/odoo/tools/convert.py", line 677, in _tag_root
    f(rec)
  File "/opt/odoo/odoo/tools/convert.py", line 580, in _tag_record
    record = model._load_records([data], self.mode == 'update')
  File "/opt/odoo/odoo/models.py", line 4220, in _load_records
    data['record']._load_records_write(data['values'])
  File "/opt/odoo/odoo/models.py", line 4149, in _load_records_write
    self.write(values)
  File "/opt/odoo/addons/payment_transfer/models/payment.py", line 68, in write
    return super(TransferPaymentAcquirer, self).write(values)
  File "/opt/odoo/addons/payment/models/payment_acquirer.py", line 335, in write
    result = super(PaymentAcquirer, self).write(vals)
  File "/mnt/data/odoo-addons-dir/server_environment/models/server_env_tech_name_mixin.py", line 66, in write
    return super().write(vals)
  File "/opt/odoo/odoo/models.py", line 3703, in write
    fields[0].determine_inverse(real_recs)
[...]

PR #111, opened 2022-05-09
[14.0][server_environment_data_encryption] Avoid raising error when config is missing

florian-dacosta:
Instead we fallback on standard server_environment behavior and display a warning to the user.

Aims to solve https://github.com/OCA/server-env/issues/110

reviews welcome @sebastienbeau @simahawk @ivantodorovich

### DIFF
--- a/server_environment_data_encryption/models/server_env_mixin.py
+++ b/server_environment_data_encryption/models/server_env_mixin.py
@@ -15,11 +15,29 @@ _logger = logging.getLogger(__name__)
 class ServerEnvMixin(models.AbstractModel):
     _inherit = "server.env.mixin"
 
+    def _current_env_encrypted_key_exists(self):
+        env = self.env["encrypted.data"]._retrieve_env()
+        key_name = "encryption_key_%s" % env
+        key_str = config.get(key_name)
+        key_exists = key_str and True or False
+        if not key_exists:
+            logging.warning(
+                "The minimal configuration is missing. You need at least to add an "
+                "encryption key for the current environment  : %s. While the "
+                "configuration is missing, the module has no effect",
+                env,
+            )
+        return key_exists
+
     def _compute_server_env_from_default(self, field_name, options):
         """First return database encrypted value then default value"""
-        self.ensure_one()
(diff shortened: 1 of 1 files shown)


ivantodorovich (approved):

simahawk (commented):
LG. Although, it's worth to add a line in the README :wink:

florian-dacosta:
> LG. Although, it's worth to add a line in the README wink

Not sure what to add.
"If the minimal configuration is not done, the module has no effect" but is seems pretty obvious no?

simahawk:
You are right. We should probably log a warning instead. What do you think?

florian-dacosta:
@simahawk 
I've added the warning

server_environment_data_encryption/models/server_env_mixin.py
```diff
@@ -15,11 +15,28 @@
...
+        env = self.env["encrypted.data"]._retrieve_env()
+        key_name = "encryption_key_%s" % env
+        key_str = config.get(key_name)
+        key_exists = key_str and True or False
+        if not key_exists:
+            logging.warning(
+                "The minimal configuration is missing. You need at least to add an "
+                "encryption key for the current environment  : {}. While the "
+                "configuration is missing, the module has no effect".format(env)
+            )
```
simahawk:
```suggestion
            logging.warning(
                "The minimal configuration is missing. You need at least to add an "
                "encryption key for the current environment: %s. While the "
                "configuration is missing, the module has no effect", env
            )
```

simahawk (commented):
can you please rewrite the commit msg?

```
s_environment_data_encryption: no error when config is missing

Fallback to normal server env behavior and log a warning.
```

florian-dacosta:
@simahawk Done

simahawk:
works on runboat:
```
2022-05-10 08:25:27,951 77 WARNING b029161ab-bf0c-4bbb-ad91-211e2b9c46d4 root: The minimal configuration is missing. You need at least to add an encryption key for the current environment  : test. While the configuration is missing, the module has no effect 
```
/ocabot merge patch

Thanks!

PR #115, opened 2022-07-01
[[14.0][REF] make server_environment_ir_config_parameter compatible with server_environment_data_encryption

florian-dacosta:
This PR in draft state and is not necessarily meant to be merged, it will probably be best to aims the v16.
But I'd like to have feedback from the users on the proposed changes and this way adapt it if there is strong blocakge about the current changes.

So, the idea is to make `ir.config_parameter` inherit from `server.env.mixin` like any other `server_environment_*` modules.
It will standardize the module which seems nice to me, and this way make it compatible with `server_environment_data_encryption` which override the `server.env.mixin` mixin.
I guess in the future we could have the same problem with any other module which aims to override `server.env.mixin` in order to add features for all `server_environment_*` modules.

The main blocking point maybe is that refactore changes the format of the configuration file.
Indeed, the current module exects something like : 
```
[ir.config_parameter]
name1=value1
name2=value2
```

While after the refactore, we have something similar as the other `server_environement_*` modules : 

```
[ir.config_parameter.name1]
value=value1

[ir.config_parameter.name1]
value=value2
```

Since this configuration is made by developpers, and since usually when you install `server_environment_ir_config_parameter` you also install other `server_environment_*` modules, in my opinion, it is not a bad thing to go with this second format because this way it won't be different from the others `server_environment_*` modules.
But, if it is a real issue, maybe we could find a way to keep the old one. (But it would complexify the module)

I'll be happy to have the opinion of the users of this module.
@sbidoul @TDu @GillesTephaneMeyomesse @legalsylvain @sebastienbeau @gurneyalex @simahawk

### DIFF
--- a/server_environment_ir_config_parameter/__init__.py
+++ b/server_environment_ir_config_parameter/__init__.py
@@ -1,1 +1,2 @@
 from . import models
+from .hook import post_init_keep_parameter_value
--- a/server_environment_ir_config_parameter/__manifest__.py
+++ b/server_environment_ir_config_parameter/__manifest__.py
@@ -10,7 +10,6 @@
     "author": "ACSONE SA/NV, Odoo Community Association (OCA)",
     "website": "https://github.com/OCA/server-env",
     "depends": ["server_environment"],
-    "data": [
-        "views/view_ir_config_parameter.xml",
-    ],
+    "post_init_hook": "post_init_keep_parameter_value",
+    "data": [],
 }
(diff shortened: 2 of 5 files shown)


florian-dacosta:
In version 16, the `_get_param` native method totally bypass the ORM and make a direct sql query.
This make `ir_config_parameter` totally incompatible with the `server.env.mixin`
This refactore is obsolete in recent version so let's close it.

PR #117, opened 2022-07-18
[15.0][ADD] mail_environment_office365

ajaniszewska-dev:

PR #119, opened 2022-08-16
[14.0][ADD] mail_environment_office365

ajaniszewska-dev:
Backport of https://github.com/OCA/server-env/pull/117


ivantodorovich (approved):

simahawk:
See comment on https://github.com/OCA/server-env/pull/117

rven:
Why not adding this one also?

mail_environment_office365/models/fetchmail_server.py

```python
from odoo import models


class FetchmailServer(models.Model):
    _inherit = "fetchmail.server"

    @property
    def _server_env_fields(self):
        base_fields = super()._server_env_fields
        office365_fields = {
            "use_microsoft_outlook_service": {},
        }
        office365_fields.update(base_fields)
        return office365_fields
```

mail_environment_office365/README.rst
```diff
@@ -0,0 +1,97 @@
...
+Configuration
+=============
+
+Example of config file ::
+
+  [ir.config_parameter]
+  microsoft_outlook_client_id = YOUR_OFFICE365_CLIENT_ID
+  microsoft_outlook_client_secret = YOUR_OFFICE365_CLIENT_SECRET
+
+  [outgoing_mail.office365_smtp_server]
```
sorvis, line 44: Will this module also cover the use case for inbound email as well?

santostelmo (approved):

JordiBForgeFlow:
I found a problem when using mail_environment in combination fetchmail_outlook, which is, as I understand, the ultimate goal of this module.

The constraint https://github.com/odoo/odoo/blob/14.0/addons/fetchmail_outlook/models/fetchmail_server.py#L32 fails, because it is unable to detect the correct value of the environment variable fields from the environment. So, it may fail due to the server_type, or is_ssl, if those are defined in the server environment. It seems to me that at the time when the constraint is called the values have not been fetched from the environment variables.

PR #127, opened 2022-10-18
[14.0] s_env_ir_config_param: allow empty values

simahawk:
Sometimes you really want an empty default value.
Make it possible on demand w/ a specific ctx key.


sebalix (approved):

petrus-v (approved):

simahawk:
I had some problems w/ this approach and I had to define a `none` value instead of leaving it empty.
It's probably due to a conflict w/ mail_environment. I have to check... but I have no time now.

PR #132, opened 2022-11-15
[14.0] server_env_ir_config_param: fix value read

damdam-s:


simahawk (changes requested):
can you please fix tests and amend the commit like this?
```
server_env_ir_config_param: fix value read

$short explanation
```
thanks :pray:

PR #134, opened 2022-12-14
[16.0][FIX] mail_environment: add search field for fetchmail.server's is_ssl

StefanRijnhart:
Fixes

```
while parsing /home/odoo/16.0/addons/mail/views/fetchmail_views.xml:88
Error while validating view near:

                    <separator/>
                    <filter string="SSL" name="ssl" domain="[('is_ssl', '=', True)]" help="If SSL required."/>
                    <separator/>
                    <filter string="Archived" name="inactive" domain="[('active', '=', False)]"/>
                </search>
Unsearchable field 'is_ssl' in path 'is_ssl' in domain of <filter name="ssl"> ([('is_ssl', '=', True)]))
```

While upgrading --all on a setup with mail_environment installed.

### DIFF
--- a/mail_environment/models/fetchmail_server.py
+++ b/mail_environment/models/fetchmail_server.py
@@ -4,6 +4,7 @@
 import operator
 
 from odoo import api, fields, models
+from odoo.osv.expression import FALSE_DOMAIN
 
 
 class FetchmailServer(models.Model):
@@ -28,6 +29,7 @@ class FetchmailServer(models.Model):
         mail_fields.update(base_fields)
         return mail_fields
 
+    is_ssl = fields.Boolean(search="_search_is_ssl")
     server_type = fields.Selection(search="_search_server_type")
 
     @api.model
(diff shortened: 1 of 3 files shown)


florian-dacosta:
thanks for the fix
There is small pre-commit issue.
About the test, I am no expert, but IMHO just a small test doing a search with the `is_ssl` field in the domain and checking the result is correct would be more than enough.

StefanRijnhart:
@florian-dacosta Thanks! Indeed, we just want to test the search method but I had to figure out how to add a server environment configuration test (and how any of this works in the first place). Now proud to add the first tests in this module!

florian-dacosta (approved):
@StefanRijnhart I think a simple "write" works to fill the value, it should work as a default value for the field if there are no server environment config.
But with a config it is fine too (maybe better so it also tests that adding a server env config works!) thanks again and LGTM

rven (approved):

rven:
Could someone merge this?

florian-dacosta:
@simahawk @etobella possible to merge this one?

simahawk (approved):

StefanRijnhart:
Pre-commit fixed in https://github.com/OCA/server-env/pull/148

PR #136, opened 2022-12-20
fix: add microsoft outlook fields hack

sorvis:


florian-dacosta:
I guess it should be in a separated module depending on `microsoft_outlook` and `mail_environment` modules.
It could be auto-installed

simahawk:
Please look at https://github.com/OCA/server-env/pull/119

PR #137, opened 2022-12-29
[14.0][FIX] README to specify server_type instead of type to match the model

sorvis:
It looks like the README config section used type, but the underlying model is actually server_type.


legalsylvain (approved):

legalsylvain:
Should be ported in the last V16 :
https://github.com/OCA/server-env/tree/16.0/mail_environment

sorvis:
> Should be ported in the last V16 : https://github.com/OCA/server-env/tree/16.0/mail_environment

Created PR:
https://github.com/OCA/server-env/pull/138

ajaniszewska-dev (approved):

PR #139, opened 2023-01-06
[FIX] server_environment: ensure _inverse_sparce is set

adrienpeiffer:
inverse method is not set on readonly field by Odoo even there is a states attributes on this field.
Here we ensure the inverse method is set if the field is editable


AnizR (commented):
Code LGTM but can you run pre-commit?

AnizR:
Can you also add a test for this particular case?

PR #141, opened 2023-01-18
[12.0][UPD] update dotfiles using copier

AungKoKoLin1997:
@qrtl


.copier-answers.yml
```diff
@@ -0,0 +1,24 @@
+# Do NOT update manually; changes here will be overwritten by Copier
+_commit: v1.14.0
```
SirTakobi, line 2: Hi! I have updated the repo configuration and fixed a couple of parameters in https://github.com/qrtl/server-env/pull/1, could you please have a look and maybe merge it? Thanks!

SirTakobi (changes requested):

.copier-answers.yml
simahawk: @SirTakobi not a blocker, you can open another PR against OCA when this is merged.

PR #146, opened 2023-03-07
[14.0] server_environment: allow setting running env through env

petrus-v:
This also give a chance to use server enivornment on odoo.sh !

### DIFF
--- a/server_environment/server_env.py
+++ b/server_environment/server_env.py
@@ -45,13 +45,20 @@ _boolean_states = {
 
 def _load_running_env():
     if not system_base_config.get("running_env"):
-        _logger.info("`running_env` not found. Using default = `test`.")
-        _logger.info(
-            "We strongly recommend against using the rc file but instead use an "
-            "explicit config file or env variable."
-        )
-        # safe default
-        system_base_config["running_env"] = "test"
+        env_running_env = os.environ.get("RUNNING_ENV", os.environ.get("ODOO_STAGE"))
+        if env_running_env:
+            system_base_config["running_env"] = env_running_env
+        else:
+            _logger.info(
+                "`running_env` or `RUNNING_ENV`, `ODOO_STAGE` not found. "
+                "Using default = `test`."
+            )
+            _logger.info(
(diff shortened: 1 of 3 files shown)


server_environment/server_env.py
```diff
@@ -52,6 +52,9 @@ def _load_running_env():
         )
         # safe default
         system_base_config["running_env"] = "test"
+    system_base_config["running_env"] = os.environ.get(
```
grindtildeath: Actually this will take precedence over what's defined in the Odoo config, with potentially different values that might be conflicting.

At camptocamp we use our docker image to make sure the Odoo config key is set by environment variable, cf: 
https://github.com/camptocamp/docker-odoo-project/blob/master/15.0/templates/odoo.cfg.tmpl#L28
https://github.com/camptocamp/docker-odoo-project/blob/master/bin/docker-entrypoint.sh#L73-L81

Anyway, if we're to have the environment variable read in this code, could we set it as a fallback to the actual Odoo config?
```
def _load_running_env():
    if not system_base_config.get("running_env"):
        env_running_env = os.environ.get("RUNNING_ENV")
        if env_running_env:
            system_base_config["running_env"] = env_running_env
        else:
            _logger.info("`running_env` not found. Using default = `test`.")
            _logger.info(
                "We strongly recommend against using the rc file but instead use an "
                "explicit config file or env variable."
            )
            # safe default
            system_base_config["running_env"] = "test"
```

petrus-v: the last commit gives the precedence to the config file

damdam-s (approved):

damdam-s:
@grindtildeath are you OK with the last changes? thanks

Koudja (approved):

simahawk:
@petrus-v could you please shorten commit msgs?

`[IMP] server_environment: allow env variable conf` should be enough

petrus-v:
@simahawk I've sqashed, rebased and reword the final commit

grindtildeath (approved):

PR #147, opened 2023-03-08
[14.0] Update pre-commit conf

petrus-v:


petrus-v:
should fix issue in #146

Issue #151, opened 2023-04-04
server_env_mixin broken: on 14.0 + 15.0 + 16.0


simahawk:
A refactoring done in odoo here https://github.com/odoo/odoo/commit/36544651f2049bcf18777091dbf02c9631b33243 which breaks server_env because the partial method cannot be inspected https://github.com/OCA/server-env/blob/14.0/server_environment/models/server_env_mixin.py#L335

Error:

```
2023-04-04 06:33:21,973 895 CRITICAL odoo odoo.service.server: Failed to initialize database `odoo`. 
Traceback (most recent call last):
  File "/opt/odoo/odoo/tools/convert.py", line 677, in _tag_root
    f(rec)
  File "/opt/odoo/odoo/tools/convert.py", line 580, in _tag_record
    record = model._load_records([data], self.mode == 'update')
  File "/opt/odoo/odoo/models.py", line 4241, in _load_records
    records = self._load_records_create([data['values'] for data in to_create])
  File "/opt/odoo/odoo/models.py", line 4157, in _load_records_create
    return self.create(values)
  File "<decorator-gen-119>", line 2, in create
  File "/opt/odoo/odoo/api.py", line 348, in _model_create_multi
    return create(self, arg)
  File "/opt/odoo-venv/lib/python3.6/site-packages/odoo/addons/component_event/models/base.py", line 102, in create
    records = super(Base, self).create(vals_list)
  File "<decorator-gen-64>", line 2, in create
  File "/opt/odoo/odoo/api.py", line 348, in _model_create_multi
    return create(self, arg)
  File "/opt/odoo/odoo/addons/base/models/ir_fields.py", line 534, in create
    recs = super().create(vals_list)
  File "<decorator-gen-13>", line 2, in create
  File "/opt/odoo/odoo/api.py", line 348, in _model_create_multi
    return create(self, arg)
  File "/opt/odoo/odoo/models.py", line 3919, in create
    fields[0].determine_inverse(batch_recs)
  File "/opt/odoo/odoo/fields.py", line 1207, in determine_inverse
    determine(self.inverse, records)
  File "/opt/odoo/odoo/fields.py", line 81, in determine
    if needle.__name__.find('__'):
AttributeError: 'functools.partial' object has no attribute '__name__'
```
Issue reported https://github.com/odoo/odoo/issues/117559

simahawk:
I'm working on a fix

rven:
[...]

PR #152, opened 2023-04-04
[FIX] AttributeError

rven:
AttributeError: 'functools.partial' object has no attribute '__name__'

### DIFF
--- a/server_environment/models/server_env_mixin.py
+++ b/server_environment/models/server_env_mixin.py
@@ -2,7 +2,7 @@
 # License LGPL-3.0 or later (https://www.gnu.org/licenses/lgpl.html)
 
 import logging
-from functools import partialmethod
+from functools import partial, update_wrapper
 
 from lxml import etree
 
@@ -332,8 +332,14 @@ class ServerEnvMixin(models.AbstractModel):
         field.compute = "_compute_server_env"
 
         inverse_method_name = "_inverse_server_env_%s" % field.name
-        inverse_method = partialmethod(type(self)._inverse_server_env, field.name)
-        setattr(type(self), inverse_method_name, inverse_method)
+        inverse_method = partial(
+            type(self)._inverse_server_env, self=self, field_name=field.name
+        )
+        setattr(
+            type(self),
+            inverse_method_name,
+            update_wrapper(inverse_method, type(self)._inverse_server_env),
+        )
         field.inverse = inverse_method_name
         field.store = False
         field.required = False


legalsylvain:
Hi @rven. It seems a duplicate of https://github.com/OCA/server-env/issues/151 Don't you think ? 
If yes, could you close your issue. 
thanks !

rven:
> Hi @rven. It seems a duplicate of #151 Don't you think ? If yes, could you close your issue. thanks !

@legalsylvain This is about the same issue, but in fact of waiting for Odoo to fix this, it can also be fixed in the way I do in this PR!
@simahawk What do you think?
This was working fine now https://github.com/OCA/storage/pull/254
image

simahawk:
I'm working on a fix on #151

server_environment/models/server_env_mixin.py
```diff
@@ -332,8 +332,14 @@ def _server_env_transform_field_to_read_from_env(self, field):
...
         inverse_method_name = "_inverse_server_env_%s" % field.name
-        inverse_method = partialmethod(type(self)._inverse_server_env, field.name)
-        setattr(type(self), inverse_method_name, inverse_method)
+        inverse_method = partial(
+            type(self)._inverse_server_env, self=self, field_name=field.name
+        )
+        setattr(
+            type(self),
+            inverse_method_name,
+            update_wrapper(inverse_method, type(self)._inverse_server_env),
```
simahawk, line 341: hmm weird... I've tried the same fix w/ update_wrapper but it didn't work...
Ah but you are not using `partialmethod` anymore.... :thinking:

simahawk: I need to double check.

simahawk (changes requested):
Just keeping on hold

server_environment/models/server_env_mixin.py
simahawk: Also, this solution won't set the right name but the name of the wrapped func. I tried overriding `__name__` but that won't work, at least on v14

rven: Works well in v16.0
image

simahawk:
@rven can you pls have a look at https://github.com/OCA/server-env/pull/153?
I don't think is correct to use `partial` instead of `partialmethod` and in any case `update_wrapper` won't propagate the right name.

rven:
> @rven can you pls have a look at #153? I don't think is correct to use `partial` instead of `partialmethod` and in any case `update_wrapper` won't propagate the right name.

It is indeed not correct, because the record(self) is also not provided well!

PR #160, opened 2023-06-19
[16.0][IMP] Server env default fields string

qgroulard:
Not all fields have their string in the args, but `.string` returns the computed (based on field name) string if not set in the args.


simahawk (approved):
LG

PR #161, opened 2023-06-20
[16.0][FIX] server_environment: string for field storing default value

glitchov:
This fix avoids a possible collision in the string of field storing default value when the first part is empty and the string is then " Env Default".


qgroulard:
Hi, I opened https://github.com/OCA/server-env/pull/160 yesterday to solve the same problem.

Regarding your solution, `fieldname` is basically `base_field.name + _env_default`, then you remove `_env_default`, thus what you use is `base_field.name `.
I propose to use `base_field.string` which looks more appropriate to me in this case.

glitchov:
@qgroulard  Indeed your solution is better than mine.

PR #163, opened 2023-08-28
[16.0][FIX] don't use deprecated `locale.getdefaultlocale()`

sbidoul:
`locale.getdefaultlocale()` is deprecated in python 3.11. Use `getlocale()` instead.

### DIFF
--- a/server_environment/system_info.py
+++ b/server_environment/system_info.py
@@ -6,6 +6,7 @@ import locale
 import os
 import platform
 import subprocess
+from functools import lru_cache
 
 from odoo import release
 from odoo.tools.config import config
@@ -19,6 +20,7 @@ def _get_output(cmd):
     return p.communicate()[0].rstrip()
 
 
+@lru_cache(maxsize=1)
 def get_server_environment():
     # inspired by server/bin/service/web_services.py
     try:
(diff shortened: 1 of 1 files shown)


sbidoul:
@simahawk @gurneyalex would you like to have a look at this little one?

While looking at this I noticed this `get_environment()` function is called repeatedly (notably each time the registry is built), and it does potentially costly subprocess calls (git, bzr, lsb_release). So I took the opportunity to cache it.

legalsylvain (approved):

legalsylvain:
> While looking at this I noticed this get_environment() function is called repeatedly (notably each time the registry is built), and it does potentially costly subprocess calls (git, bzr, lsb_release). So I took the opportunity to cache it.

AFAIK, get_server_environment is called only by the model ``server.config`` that has, In my opinion nothing to do in this module. It allows to see configuration parameter in a view of a transient model. 

image


1) this feature can be usefull without the module server_environment installed.
2) something may be wrong in this part of the code, if you notice that calls are made regularly. (this should not be the case).
3) this feature can provide security issues.


So I'm in favor to move all that part (``server.config``  + system_info.py) in a dedicated module named ``server_configuration_visualisation`` (or something like this).

What do you think ? 

Note : this code didn't changed since the creation (version 6.1 : https://github.com/OCA/server-tools/blob/6.1/server_environment/system_info.py)

lmignon (approved):

rousseldenis (approved):

PR #166, opened 2023-09-25
[14] server_environment: fix default method not passed

hparfr:
before this commit, default values were not populated in other environments. The problem was only seen with data_encryption which save all the values for all the environments.


hparfr:
ping @florian-dacosta @sebastienbeau

sebastienbeau (approved):
LGTM (code review)

good catch !

simahawk (approved):

PR #172, opened 2023-12-01
[FIX] Allow manual configuration of incoming mail server

rven:
When manual configuring an incoming mail server, the changes are not saved.

The problem is because the fields are marked as readonly by the server environment module.

With this change it is allowed to manually add an incoming mail server.

### DIFF
--- a/server_environment/models/server_env_mixin.py
+++ b/server_environment/models/server_env_mixin.py
@@ -404,6 +404,7 @@ class ServerEnvMixin(models.AbstractModel):
             fieldlabel = "{} {}".format(base_field.string or "", "Env Default")
             field_args.update(
                 {
+                    "readonly": False,
                     "sparse": "server_env_defaults",
                     "automatic": True,
                     "string": fieldlabel,


len-foss:
What is the problem to be solved? 

Improper configuration of email server in an environment is fairly risky, and so it is a feature that it would be impossible to change manually. Making this patch the default would mean losing the safety this brings. And changing something like that without big warnings would create problems for sure.

If the argument is that it is not a big problem if backup/restoration scripts make sure that there are no such risks, then why would this feature be needed?

My understanding would be that this change should be in another optional module if you really want it, but the default behaviour shouldn't change.

rven:
With this change it is possible to create manual entries in the mail configuration.

The mail servers defined in the configuration are still not updateable!

It's a bit strange to that you suggest to create an extra module to enable standard Odoo functionality again?

Outgoing mail servers can be entered manually!

BTW if you neutralize the DB it is done automatically by Odoo parts/odoo/addons/mail/data/neutralize.sql

len-foss:
In that case, why bother with `mail_environment` at all? You can just uninstall the module and put mail environment changes into your backup/restore scripts, and then you can change everything manually.

rven:
We use them in combination.
The "test" environments are defined in the configuration to enable mailtrap by default.
But the customer should be able to change / add incoming mail servers manually in the production environment.
I don't see the problem why this change is breaking stuff! It just fixes a "bug" in this module.

len-foss:
In that case it seems the need is covered by `server_environment_data_encryption` as is discussed in more details in issue 169.

rven:
Does this also allow to manually create an incoming mail server?
Because the values are not accepted at all, because the fields are readonly...

PR #177, opened 2024-02-14
[15.0] auth_saml_environment: removes readonly attributes to allow provider creation

StephaneMangin:
Some readonly fields prevent the creation of a provider. Removing them as they will be managed by the env.

Following change in https://github.com/OCA/server-auth/pull/342

(edit @yvaucher adding link to the source of the issue)

### DIFF
--- a/auth_saml_environment/models/auth_saml_provider.py
+++ b/auth_saml_environment/models/auth_saml_provider.py
@@ -8,6 +8,11 @@ class AuthSamlProvider(models.Model):
     _name = "auth.saml.provider"
     _inherit = ["auth.saml.provider", "server.env.mixin"]
 
+    # Mandatory to be able to create objects
+    idp_metadata = fields.Text(required=False)
+    sp_pem_public = fields.Char(required=False)
+    sp_pem_private = fields.Char(required=False)
+
     sp_pem_public_path = fields.Char(
         string="sp_pem_public_path env config value",
     )


auth_saml_environment/models/auth_saml_provider.py
```diff
@@ -8,6 +8,11 @@ class AuthSamlProvider(models.Model):
     _name = "auth.saml.provider"
     _inherit = ["auth.saml.provider", "server.env.mixin"]
 
+    # Mandatory to be able to create objects
```
yvaucher, line 11:
```suggestion
    # Non-mandatory to be able to create objects
```

yvaucher:
You don't want to state the odoo version in the commit message but rather the type of change. The odoo version is to be mentionned in the title of the PR.

Can you replace the content of the commit message by:

```
    [FIX] auth_saml_environment: no require on readonly fields
    
    Restore the usage of the form view by dropping the required condition on
    binary fields that are readonly.
    
    Those fields became mandatory after https://github.com/OCA/server-auth/pull/342
    but the module auth_saml_environment offer to use paths as a alternative config
    option for binaries thus those fields can be empty.

```

yvaucher:
Backported here https://github.com/OCA/server-env/pull/179

vincent-hatakeyama:
idp_metadata is required for a provider to work. In a similar way, the server hostname of an outgoing server is mandatory.  mail_environment does not change smtp_host to be optional, so I do not see why this module would do so for idp_metadata.

For sp_pem_public and and sp_pem_private, I would recommend fixing issue OCA/server-auth/issues/315 in auth_saml that made me change them to mandatory rather than adding another level of work around.

PR #188, opened 2024-06-21
[17.0][ADD] server_environment_autocreate: add possibility of auto creating records

vincent-hatakeyama:
We need to create mail server or fs.storage in the interface to match those configured with server_environment. This enhancement proposes to let server_environment creates it automatically instead.


vincent-hatakeyama:
@OCA/server-environment-maintainers Reviews welcome

server_environment/models/server_env_mixin.py
```diff
@@ -424,3 +425,40 @@ def _setup_base(self):
...
+                                    [
+                                        {
+                                            "name": section,
+                                            "model": model_name,
+                                            "module": "__server_environment__",
+                                            "res_id": record.id,
+                                        }
+                                        for record in records
+                                    ]
+                                )
```
simahawk: Why would you loop on all the models? This mixin will be inherited by a specific model and that model only should be set up... 
```suggestion
        global_section_name = self._server_env_global_section_name()
        for section in serv_config:
            if not section.startswith(f"{global_section_name}."):
                continue
            if not serv_config[section].get("__autocreate", None):
                continue
            name_value = section[len(global_section_name) + 1 :]
            domain = [
                (model._server_env_section_name_field, "=", name_value)
            ]
            count = model.with_context(active_test=False).search_count(
                domain
            )
            if not count:
                _logger.info("Automatic creation of %s", section)
                values = literal_eval(
                    serv_config[section].get("__autocreate")
                )
                values[
                    model._server_env_section_name_field
                ] = name_value
                # TODO: handle failure
                records = model.create([values])
                self.env["ir.model.data"].create(
                    [
                        {
                            "name": section,
                            "model": model_name,
                            "module": "__server_environment__",
                            "res_id": record.id,
                        }
                        for record in records
                    ]
                )
```
Something like this would be better.

simahawk (changes requested):
Hello, thanks for your contrib.

TBH Is not a feature that I would include in our projs but can be interesting for others.

However:
* you should move this feature to a separate module
* you should take care of failure: some create can fail (eg: missing required fields)
* it can be very tricky to handle because you might have models that are not fully loaded yet

vincent-hatakeyama:
@simahawk 

> you should move this feature to a separate module

This is a feature that is optional, so I did not see the need for a separate module. Have you any reason for doing it in a separate module?

Concerning your other point, I do not expect all cases to work out of the box. Basic cases (outgoing mail servers, fs storage), it works and the current version is already in production.

simahawk:
> This is a feature that is optional

That's exactly the main reason to move it to its own module. This is not something we want to maintain OOTB ;)

vincent-hatakeyama:
I changed the feature to a new module as suggested.
I also added more tests and documentation of the feature.

server_environment_autocreate/models/server_env_mixin.py
```diff
@@ -0,0 +1,59 @@
...
+
+class ServerEnvMixin(models.AbstractModel):
+    """Inherited to add autocreation of records when register_hook is called."""
+
+    _inherit = "server.env.mixin"
+
+    _server_env_allow_autocreate: bool = True
+
+    @api.model
+    def _reverse_server_env_section_name(self, section_name: str) -> None | str:
```
simahawk: Beside is hard to understand what is the purpose of this method, IMO this should be named `_get_server_env_section_name`.
The docstring should say what it does 1st then explain that has to be customized under certain circumstances.

server_environment_autocreate/pyproject.toml
```diff
@@ -0,0 +1,6 @@
+[build-system]
+requires = ["whool"]
+build-backend = "whool.buildapi"
+
+#[project.optional-dependencies]
```
simahawk: While I like this idea, it should be removed to avoid future conflicts.
However, you could propose this improvement to the whool lib https://github.com/sbidoul/whool

simahawk (commented):
LG overall, minor comments

server_environment_autocreate/pyproject.toml
vincent-hatakeyama: Opened sbidoul/whool#26

server_environment_autocreate/models/server_env_mixin.py
vincent-hatakeyama: Updated.

server_environment_autocreate/pyproject.toml
vincent-hatakeyama: Comment removed.

server_environment/readme/ROADMAP.md
```diff
@@ -1,5 +1,5 @@
 - it is not possible to set the environment from the command line. A
-  configuration file must be used.
```
simahawk: these changes to server_env are not related. If you want to keep them here, pls move them to their own commit

simahawk (commented):
Almost there :)

server_environment/readme/ROADMAP.md
vincent-hatakeyama: I’ve removed them.

Created #231 for the fix on the import.

simahawk (approved):

PR #218, opened 2024-12-17
[FIX] server_environment: Fix tests on complex environments

etobella:
Without this change, the test will fail if we have queue_job installed (for example)


LoisRForgeFlow (approved):
LGTM

luisDIXMIT (approved):
LGTM!

PR #219, opened 2025-01-02
[IMP] mail_environment: Keep existing settings at install of module

rven:
The Outgoing / Incoming mail servers configuration in the Odoo environment, will get lost when the mail_environment module gets installed.

With this change, a post installation hook is added which will add the current values to the server_env_defaults.
So the currently configured and working mail servers will still be working after the installation of this module.


gurneyalex (approved):

rven:
Still relevant

petrus-v (approved):
Nice one ! thanks

gurneyalex (approved):

PR #224, opened 2025-03-31
[18][MIG] data_encryption

florian-dacosta:
@bguillot @hparfr


rvalyi (approved):

bguillot (approved):

data_encryption/models/encrypted_data.py
```diff
@@ -79,35 +80,33 @@ def _encrypted_read_json(self, name, env=None):
...
         """
-        key_name = "encryption_key_%s" % env
+        key_name = f"encryption_key_{env}"
         key_str = config.get(key_name)
         if not key_str:
             raise ValidationError(
-                _(
+                self.env._(
                     "No '%(key_name)s' entry found in config file. "
                     "Use a key similar to: %(key)s"
```
hparfr: how key_name is injected into this ?
where key is defined ?

hparfr (changes requested):

data_encryption/models/encrypted_data.py
florian-dacosta: I am not sure to understand what you mean.
Key_name and key are injected 2 lines below

hparfr: 1) injection: I don't remember the good syntax; but usualy is in the form ` f"{key_name}"` or `fun("%(key_name)s", key_name=key_name)`, or `"%(key_name)s".format(key_name=key_name)` ... 
2) `key`: I see no definition of variable `key` in `_get_cipher()`

florian-dacosta: 1. I never know the "good way" ttto inject variable into strings, there are 10 ways to do it...
2. `key` is defined to lines below, when injecting the variable `"key": Fernet.generate_key()`

I can change the syntax with the `"my string".format() `one, is this what you suggest ?
`fstring` should not be used for translated string right ?

hparfr: 1) yup.  you may try `_("%(key_name)s", key_name=key_name) ` looks `_` accepts more than one arg https://www.odoo.com/documentation/18.0/developer/howtos/translations.html?highlight=translation#variables

2) Good !

hparfr (approved):
Thanks !

PR #226, opened 2025-04-01
[18][MIG] server_environment_data_encryption

florian-dacosta:
Standard migration
There is some changes in the module because it seems that custom context (`environment` key here) are not passed to `get_views` method. 
Because of this, we can't manage a different key for the ormcache to manage 1 different view per environment.

So the strategy here is to build a single view with all environment buttons/alert with `invisible` attributes depending on the context.


bguillot (approved):

bguillot:
Code review + test

hparfr (approved):
Tested with mail_environment, works as expected

rvalyi (approved):

PR #231, opened 2025-05-16
[17.0][FIX] server_environment: use local import

vincent-hatakeyama:


server_environment/static/description/index.html
```diff
@@ -543,11 +543,11 @@ <h1><a class="toc-backref" href="#toc-entry-8">Known issues / Roadmap</a></h1>
 <ul class="simple">
 <li>it is not possible to set the environment from the command line. A
 configuration file must be used.</li>
-<li>the module does not allow to set low level attributes such as database
-server, etc.</li>
+<li>the module does not allow to set low level attributes such as
+database server, etc.</li>
```
simahawk, line 547: FYI if you want to disable readme update when not needed, you can do `export SKIP=oca-gen-addon-readme` :wink:

simahawk (approved):

server_environment/static/description/index.html
vincent-hatakeyama: Thanks for the tip.

PR #234, opened 2025-06-04
[18.0] Fix inheritance of server env techname mixin

grindtildeath:
In case a model was inheriting both server.env.mixin and server.env.techname.mixin the value of _server_env_section_name_field was depending on the order in which the mixins were inherited on that model.

If server.env.mixin was inherit first as in:
_name = "my.model"
_inherit = ["my.model", "server.env.mixin", "server.env.techname.mixin"]

Then the value of _server_env_section_name would be `name`

If server.env.techname.mixin was inherit first as in: _name = "my.model"
_inherit = ["my.model", "server.env.techname.mixin", "server.env.mixin"]

Then the value of _server_env_section_name would be `tech_name`

To avoid this, since the server.env.techname.mixin changes an attribute of server.env.mixin, server.env.techname.mixin must inherit the server.env.mixin.

With this, only one of the mixins needs to be inherited and there cannot be any 'silent' issue, where the _server_env_section_name depends on the order on which mixins are inherited.

FW Port:
 - [ ] #233


server_environment/README.rst
```diff
@@ -100,12 +96,12 @@ You can edit the settings you need in the ``server_environment_files``
 addon. The ``server_environment_files_sample`` can be used as an
 example:
 
-- values common to all / most environments can be stored in the
-  ``default/`` directory using the .ini file syntax;
-- each environment you need to define is stored in its own directory and
-  can override or extend default values;
-- you can override or extend values in the main configuration file of
-  your instance;
+-  values common to all / most environments can be stored in the
```
simahawk, line 99: next time, use `export SKIP=oca-gen-addon-readme` to disable readme updates ;)

simahawk (approved):

PR #227, opened 2025-04-07
[18.0][FIX] server_environment: set env-computed fields to readonly

rven:
Readonly fields are not calculated correctly because `_fields_view_get` was not replaced by `_get_view` during the migration

PR #236, opened 2025-06-23
[FIX] server_environment_ir_config_parameter: Don't save values in DB

rven:
When a system parameter doesn't exist on the production environment, the parameter is created with the environment value that is defined on the production system.

When this DB is exported to a local or other environment, where the environment value is NOT defined on the system, the DB uses the one that has been saved in the production DB.
It is totally unnecessary to store them in the DB, because the get_param function always gets them from the environment, when defined.
Also when you define them in a data file with following values:
```xml
    <record model="ir.config_parameter" id="ir_config_parameter_test">
        <field name="key">test</field>
        <field name="value">PLACEHOLDER</field>
    </record>
```
and the key is defined in the environment that one will also be used to create the system parameter, instead of the one defined in the data file.

There is logic in place that protects server environment defined fields from being updated at all, see https://github.com/OCA/server-env/pull/227

There is no need at all to store these values in the DB!

Everything also works fine when the system parameter is not defined at all.
It will not appear in the list view of the system parameters, but is viewable with the "Server Environment" action.

### DIFF
--- a/server_environment_ir_config_parameter/models/ir_config_parameter.py
+++ b/server_environment_ir_config_parameter/models/ir_config_parameter.py
@@ -34,42 +34,16 @@ class IrConfigParameter(models.Model):
                     _("Key %s is empty in " "server_environment_file") % (key,)
                 )
             if cvalue != value:
-                # we write in db on first access;
-                # should we have preloaded values in database at,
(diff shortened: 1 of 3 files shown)


rven:
no stale

PR #240, opened 2025-07-04
[18.0][ADD] server_environment: Add migration script

mmequignon:
In 17.0, default fields were renamed `x_{field_name}_env_default`.
Those fields are referenced in the `server_env_defaults` sparse column, but the keys here needs to be updated.
Adding a migration script to fix those values.


sebalix:
Closing in favor of https://github.com/OCA/server-env/pull/241

PR #239, opened 2025-07-02
[FW][17.0] [IMP] mail_environment: Add the `smtp_authentication` field to the `ir.mail_server` model

ivantodorovich:
Forward-port of:
- https://github.com/OCA/server-env/pull/237

PR #243, opened 2025-08-04
[18.0][MIG] mail_environment_office365: Migration to 18.0

yankinmax:


yankinmax:
@gurneyalex can you look? It seems good

mail_environment_office365/models/ir_mail_server.py
```diff
@@ -0,0 +1,18 @@
...
+
+
+class IrMailServer(models.Model):
+    _inherit = "ir.mail_server"
+
+    @property
+    def _server_env_fields(self):
+        base_fields = super()._server_env_fields
+        office365_fields = {
+            "smtp_authentication": {},
```
ivantodorovich, line 15: We should merge and forward port this instead, IMO:
- https://github.com/OCA/server-env/pull/239

The `smtp_authentication` field is not specific to office365.


----

Is there no other field needed to be set through the environment, for outlook?
Like the `microsoft_outlook_access_token` or `microsoft_outlook_refresh_token` ?

ivantodorovich (changes requested):

mail_environment_office365/models/ir_mail_server.py
yankinmax: I'm not sure @ivantodorovich 
This module is a specific one. It provides a description of the parameters required for the outlook.
If we go with forward port we will need to update the description on `mail_environment`.
Maybe it's not obvious for everyone in the community.

ivantodorovich: What I'm sure of is that the `smtp_authentication` field is defined in the base `mail` module, and re-used by `microsoft_outlook` and `google_gmail` (and possibly others) -- so making it configurable through environment belongs to `mail_environment`, not here.

However, I think you need to make other fields configurable through env, that are truly specific to `microsoft_outlook`, like the ones I suggested above. But to know this you need to test and see if you can fully configure an email server with outlook credentials by playing solely with the environment

yankinmax:
This PR will be closed in favor for:
- [ ] https://github.com/OCA/server-env/pull/244

PR #247, opened 2025-10-02
[19.0][MIG] server_environment: Migration to 19.0 

lembregtse:
We still want to improve to create a seperate config section for running_env.


server_environment/security/res_groups.xml
```diff
@@ -7,7 +7,7 @@
     <record model="res.groups" id="has_server_configuration_access">
         <field name="name">View Server Environment Configuration</field>
         <field
-            name="users"
+            name="user_ids"
             eval="[(4, ref('base.user_root')), (4, ref('base.user_admin'))]"
```
FrancoMaxime: Since Odoo 16, we can use Command instead of magic number
`eval="[Command.link(ref('base.user_root')), Command.link(ref('base.user_admin'))]"`

server_environment/server_env.py
```diff
@@ -173,16 +173,14 @@ class ServerConfiguration(models.TransientModel):
...
 
         """
-        ModelClass = super()._build_model(pool, cr)
-        ModelClass._add_columns()
-        ModelClass._arch = None
-        ModelClass._build_osv()
-        return ModelClass
+        cls._add_columns()
+        cls._arch = None
+        cls._build_osv()
```
FrancoMaxime, line 183: Since _build_model doesn't exist anymore in odoo.BaseModel, how this method is triggered now?

server_environment/security/res_groups.xml
dynapps: I will adapt this, thanks.

server_environment/server_env.py
dynapps: This is called through a monkey patch of the function in model_classes (add_to_registry) that they now use to instantiate all these classes, of course open to alternative suggestions.

FrancoMaxime: Oh thanks for the explanation, i missed it

FrancoMaxime (approved):
LGTM: code & functional review

cuongnmtm:
Hi, do you have a way to remove this new warning from Odoo?

`2025-10-02 12:59:43,786 32170 WARNING ? odoo.tools.config: unknown option 'running_env' in the config file at ...{project}/config/odoo.conf, option stored as-is, without parsing`

FrancoMaxime:
> Hi, do you have a way to remove this new warning from Odoo?
> 
> `2025-10-02 12:59:43,786 32170 WARNING ? odoo.tools.config: unknown option 'running_env' in the config file at ...{project}/config/odoo.conf, option stored as-is, without parsing`

I'm still looking for a solution too

sbidoul:
@FrancoMaxime in environments where we set the RUNNING_ENV variable, the running_env option in odoo.cfg is redundant, so we can remove it? https://github.com/acsone/odoo-bedrock/pull/59

sbidoul:
In `server_env_mixin` there is also an override of `_setup_base` that has been removed from upstream too? Does that need adapting? 

I certainly hope we can find a solution without monkey patching.

update: I don't understand how it could work without that override and why the tests pass. I may be overlooking something.

cuongnmtm:
> @FrancoMaxime in environments where we set the RUNNING_ENV variable, the running_env option in odoo.cfg is redundant, so we can remove it? [acsone/odoo-bedrock#59](https://github.com/acsone/odoo-bedrock/pull/59)

@sbidoul  From what I understand, setting the running_env option in odoo.cfg defines the environment per instance rather than for the whole server. Is that correct, or is there another recommended way to handle this, such as setting environment variables directly on the Odoo process?

cuongnmtm:
I'm new to these modules and still trying to wrap my head around how it all works, so I hope it's fine to ask here. I noticed the README mentions server_environment_files and server_environment_files_sample, but I haven’t been able to locate them. Could you point me in the right direction?

sbidoul:
@cuongnmtm you can definitely set en environment variable per Odoo instance, depending on how you start the service.
`server_environment_files_sample` has likely not been migrated. Have a look in previous branches.

sbidoul (dismissed):
Need investigation of the `_setup_base` override.

sbidoul:
@lembregtse I would suggest doing the running env section part in another PR, as it is not strictly necessary for migration, so we keep this migration PR reasonably small.

dynapps:
I think for _setup_base we can use the inheritable _post_model_setup__ which is now called in 19.0 from setup_model_classes, which is called by the registry somewhere after _setup (equivalent from _setup_base).

As for the _build_model I currently did not find a viable path to achieve the same functionality without monkeypatching. But I'm open to suggestions.

dynapps:
> @lembregtse I would suggest doing the running env section part in another PR, as it is not strictly necessary for migration, so we keep this migration PR reasonably small.

I'll change it, thanks.

rven:
I added some tests, because functionality in this module was not fully verified.

rven:
Also incorporated changes already proposed in v18 https://github.com/OCA/server-env/pull/227

cyrilmanuel (approved):

cyrilmanuel:
hi @lembregtse thanks for your work ! 
hi @sbidoul  is it good for you  ? can we merge this PR please :)

rousseldenis:
@sbidoul

sbidoul (commented):
Looks good to me.

I'm still slightly annoyed with the monkey patch. As we can't get rid of it, I wonder if it's not the right time to remove the `server.config` model and move it to another module. It is only used for display. I don't know if many people view the server config through that UI. I personally don't and I always worried about the security implications of exposing server environment through Odoo RPC.

cyrilmanuel:
@sbidoul  @lembregtse , hi guys, thanks for this pr, could we fix it or accepte to deploy as it ?

sbidoul:
@simahawk @gurneyalex as I know you are heavy users of server_environment, what do you think about removing the UI part, and eventually moving it to another module if anyone needs it? It is complex, and I often wonder if it's not a security risk.

leemannd:
Hello @sbidoul , thank you for the reminder. We are currently thinking about a better solution.
I think that we could merge the PR and then come back with improvements.

I will do an internal followup of the question.

cyrilmanuel:
seems to be done here ? could we merge these 2 PR ? https://github.com/OCA/server-env/pull/249

lmignon (approved):

sbidoul:
My preference goes to removing the UI part (and possibly move it to another module if anyone needs it), but that is not a blocker for me.

lmignon:
As this PR blocks a number of other PRs, I propose merging it in the absence of any response to Stéphane's latest proposals, so that the other PRs can be merged/released in turn.

simahawk:
> @simahawk @gurneyalex as I know you are heavy users of server_environment, what do you think about removing the UI part, and eventually moving it to another module if anyone needs it? It is complex, and I often wonder if it's not a security risk.

@sbidoul sorry, I've completely missed your ping :sweat_smile: 

Yes, I agree w/ your proposal, however we are currently working on a solution to get rid of server env.
We will post something here when we have a working prototype :wink:

PR #251, opened 2025-11-17
[14.0][IMP] server_environment_data_encryption: support binary fields

hparfr:
Encode binary fields (bytes) as b64


florian-dacosta (approved):
Nice addition
Could you eventually port it to 18 please ?

PR #257, opened 2026-02-24
[18.0][FIX] server_environment: fix excessive logging noise when running tests

florentx:
Related to:
- odoo/odoo#247151

1. Fix excessive logging
2. Add test cleanup, to pass the new check

### DIFF
--- a/server_environment/models/server_env_mixin.py
+++ b/server_environment/models/server_env_mixin.py
@@ -7,6 +7,7 @@ from functools import partialmethod
 from lxml import etree
 
 from odoo import api, fields, models
+from odoo.tools import mute_logger
 
 from odoo.addons.base_sparse_field.models.fields import Serialized
 
@@ -356,7 +357,9 @@ class ServerEnvMixin(models.AbstractModel):
         inverse_method = _partialmethod(
             type(self)._inverse_server_env, field.name, __name__=inverse_method_name
         )
-        setattr(type(self), inverse_method_name, inverse_method)
+        # Mute message related to new safeguard (PR odoo/odoo#247151)
+        with mute_logger("odoo.tests.common"):
+            setattr(type(self), inverse_method_name, inverse_method)
         field.inverse = inverse_method_name
         field.store = False
         field.required = False
--- a/server_environment_data_encryption/tests/test_server_environment_data_encrypt.py
+++ b/server_environment_data_encryption/tests/test_server_environment_data_encrypt.py
@@ -14,18 +14,27 @@ class TestServerEnvDataEncrypted(CommonDataEncrypted):
         super().setUpClass()
         cls.loader = FakeModelLoader(cls.env, cls.__module__)
         cls.loader.backup_registry()
+        cls._origin_fields = {}
+        for model in "res.partner", "res.users":
+            cls._origin_fields[model] = set(dir(cls.env[model].__class__))
 
+    def setUp(self):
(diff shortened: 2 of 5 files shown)


server_environment_data_encryption/tests/test_server_environment_data_encrypt.py
```diff
@@ -14,18 +14,27 @@ def setUpClass(cls):
...
+        self.set_new_key_env("prod")
+        self.set_new_key_env("preprod")
+        self.addCleanup(self.loader.restore_registry)
+        self.addCleanup(self.remove_mixin_fields)
 
-    @classmethod
-    def tearDownClass(cls):
-        cls.loader.restore_registry()
-        super().tearDownClass()
+    def remove_mixin_fields(self):
```
simahawk, line 32: do we need this if we update odoo-test-helper and move the setup to setUp?

florentx: I've tested against the latest release of `odoo-test-helper`  (2.1.3, February 20th).
This change was still needed.

yvaucher (approved):

yvaucher:
Can you squash your commits?

yvaucher:
@simahawk Can you finish your review? IIUC this should fix the tests and unlock other PRs.

acsonefho (approved):

StephaneMangin (approved):
LGTM

server_environment/models/server_env_mixin.py
```diff
@@ -356,7 +356,8 @@ def _server_env_transform_field_to_read_from_env(self, field):
         inverse_method = _partialmethod(
             type(self)._inverse_server_env, field.name, __name__=inverse_method_name
         )
-        setattr(type(self), inverse_method_name, inverse_method)
+        # Do not use 'setattr', to skip safeguard (PR odoo/odoo#247151)
+        type.__setattr__(type(self), inverse_method_name, inverse_method)
```
sbidoul: Would a mute_logger like in https://github.com/OCA/server-auth/pull/899 work too?

petrus-v (approved):

server_environment/models/server_env_mixin.py
florentx: yes, it's other possibility:
```python
    with mute_logger("odoo.tests.common"):
        ...
```

I'm slightly in favor of this change here, since it avoids having too much noise in the code which is only for unit test.

sbidoul: If that works and we can avoid the `__setattr__` I'd prefer that too.

simahawk: I find that approach awkward to be honest 🤔

florentx: I had a preference for `type.__setattr__` , but I can switch to `mute_logger` if it is the consensus.

sbidoul (approved):

server_environment/models/server_env_mixin.py
sbidoul: To elaborate, I find that calling dunder methods generally raises an alarm in my head.
What's still unclear to me is why Odoo test mode insists on logging on that. I understand the error if a test does not cleanup after itself, but calling setattr in itself has nothing wrong when one knows what one is doing.

victoralmau (approved):

Issue #258, opened 2026-03-03
RFC: replacement of server_environment


gurneyalex:
# RFC: replacement of `server_environment`

Authors:
* @ivantodorovich 
* @simahawk 

## Issues with server_environment we would like to fix

* Reduce maintenance work: no more glue modules for every little field to be configured through env

* Allow non-technical people to easily configure things directly in Odoo UI (e.g: test new mail server, sftp, etc)

* Drop the hacky conversion to computed fields

    * It breaks the CLI neutralize queries (used in our platform and in Odoo.sh)

    * It can break core bussiness logic due to computed fields behaving differently

## Explored alternatives

1. **Replacement with a module that encrypts fields directly in db**
    Too complicated. Even more complicated than the current solution.
2. **Explore alternative without computed fields, like fetch**
    Still hacky and still requires lots of maintenance.
3. **Replacement with an SQL-based entrypoint**
    Chosen solution. Described below.

 
## Proposed solution

Inspired by the neutralize command, that executes raw SQL queries to clear credentials out from the database, we propose building a **command line tool that writes environment-specific field values directly to the db**. Essentially, this means preparing the database for the environment that’s gonna run on, and letting Odoo work on its own without any hacks.

The tool will only overwrite the credentials in the spec, but it will not worry about clearing them out. For that, we will rely directly on the neutralize command and we will encourage developers to include neutralization scripts on the modules that require them.

The idea is then to:

*  Configure production environments directly from the UI and stored in the db – purely standard, no added magic

*  Rely on the neutralization script to clear them out when restoring a db to a non-production environment.

* Use the new tool to transform an input file/content set through the environment to SQL-queries, and set the environment-specific settings/credentials

This will be used to, for example, connect to a test sftp server instead. It will not be needed if we don’t want the test instance to connect to a specific server, though, as we will rely on neutralize to have it disabled anyway.

 
### Packaging

This will be an external tool, not an Odoo module, that will work across Odoo versions.

 
### Supported Odoo Versions

It must support Odoo >= 16.0, which is when neutralize was implemented.

> The tool itself could be used to neutralize a list of model/fields by setting them to a _falsy_ value, and so support previous versions too.

 
### Input

The tool input will be a spec that defines the records and fields values to overwrite.

Originally we wanted to support a syntax similar to the one in the existing `server_environment` module: the ini format configuration files, using sections for model-wide settings and dotted notation for record-specific settings.

[...]

PR #264, opened 2026-04-02
[18.0][IMP] server_environment: Preserve not env managed data

qgroulard:
Heavily inspired from https://github.com/OCA/server-env/pull/219, thank you @rven for your contribution.

I wanted to make @rven script more generic because this problem potentially happens every time we add the `server.env.mixin` to a model or when we add fields to the `_server_env_fields()` list.

------------------------------------------

Helper function typically used for hooks and migration scripts.
Restores database values for fields transitioning to 'server env managed'.

When a field is defined as managed by the server environment, Odoo ignores the value stored in the database, prioritizing the environment configuration instead. If no environment configuration exists, the field may effectively lose its previous value.

This method forces to 'persist' these values if they are not explicitly overridden by the current environment configuration.

### DIFF
--- a/server_environment/models/server_env_mixin.py
+++ b/server_environment/models/server_env_mixin.py
@@ -428,3 +428,31 @@ class ServerEnvMixin(models.AbstractModel):
             self._server_env_transform_field_to_read_from_env(field)
             self._server_env_add_is_editable_field(field)
         return
+
+    @api.model
+    def _preserve_not_env_managed_data(self, field_name_list):
+        """
+        Helper function typically used for hooks and migration scripts.
+        Restores database values for fields transitioning to 'server env managed'.
+
+        When a field is defined as managed by the server environment, Odoo
+        ignores the value stored in the database, prioritizing the environment
+        configuration instead. If no environment configuration exists, the field
+        may effectively lose its previous value.
+
+        This method forces to 'persist' these values if they are not
+        explicitly overridden by the current environment configuration.
+        """
+        self.env.cr.execute(f"SELECT * FROM {self._table}")
+        for row in self.env.cr.dictfetchall():
+            record = (
+                self.env[self._name]
+                .with_context(active_test=False)
+                .search([("id", "=", row["id"])])
+            )
+            if record:
+                record_values = {}
+                for field_name in field_name_list:
+                    if field_name in row:
+                        record_values[field_name] = row[field_name]
+                record.update(record_values)
--- a/mail_environment/migrations/18.0.1.0.1/post-migration.py
+++ b/mail_environment/migrations/18.0.1.0.1/post-migration.py
@@ -1,0 +1,9 @@
+# Copyright 2026 ACSONE SA/NV
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from openupgradelib import openupgrade
+
+
+@openupgrade.migrate()
+def migrate(env, version):
+    env["ir.mail_server"]._preserve_not_env_managed_data(["smtp_authentication"])


server_environment/models/server_env_mixin.py
```diff
@@ -428,3 +428,33 @@ def _setup_base(self):
             self._server_env_transform_field_to_read_from_env(field)
             self._server_env_add_is_editable_field(field)
         return
+
+    @api.model
+    def preserve_not_env_managed_data(self):
```
sbidoul:
```suggestion
    def _preserve_not_env_managed_data(self):
```

If this is to be used in migration scripts, maybe it's safer to pass the list of fields/columns to handle explicitly?

rousseldenis (approved):
LGTM

rven (approved):

server_environment/models/server_env_mixin.py
```diff
@@ -428,3 +428,31 @@ def _setup_base(self):
...
+            record = (
+                self.env[self._name]
+                .with_context(active_test=False)
+                .search([("id", "=", row["id"])])
+            )
+            if record:
+                record_values = {}
+                for field_name in field_name_list:
+                    if field_name in row:
+                        record_values[field_name] = row[field_name]
```
sbidoul, line 457: To be complete, this may need conversion based on the field type. But this is good enough for practical use cases of server env fields.

yankinmax:
I'm working on the issue:
- [ ] https://github.com/OCA/server-env/issues/258

Therefore, I've opened several PR's to remove `server_environment` dependency from `OCA/storage` and `OCA/web-api` modules.

For example:
- [ ] https://github.com/OCA/storage/pull/655

Only on that **v18 split PR** the **current PR** issue was raised in the tests and the logical fix was to add a `_preserve_not_env_managed_data` post init hook on the impacted modules.

Why do we need this hook there:

Before this split (`storage_backend` to `storage_backend` and `storage_backend_environment`), configuration fields such as `storage.backend::backend_type` and `storage.backend::filename_strategy` were normal stored database fields.
The new environment module apply `server.env.mixin` to these fields. The mixin changes how their values are obtained:
1. If an environment key exists, its value is used.
2. Otherwise, a value stored in `server_env_defaults` is used as a fallback.

When the environment module is installed on an existing database, the old value is still present in the original SQL column, but it is not automatically copied into `server_env_defaults`.
As a result, Odoo can return `False` instead of the existing value.
This caused errors such as component usage `False` and empty filename strategies.
`storage_backend_environment` module therefore will have a `post_init_hook` calling `_preserve_not_env_managed_data`.
The helper reads the original column values and stores them as environment fallbacks after the module is installed.
The preserved database value is used only when the corresponding environment key is absent.

The question which now comes to me:
1. Do we need to backport this helper to v16 and v17, forward port to v19?
2. Should we reuse it on each branch starting from v16?

From what I see it impacts all the versions.

As for now we've decided to cover `server_environment` dependency removal starting from v16:

`OCA/storage` PR's that can be updated with `_preserve_not_env_managed_data` post init hook:

- [ ] https://github.com/OCA/storage/pull/653
- [ ] https://github.com/OCA/storage/pull/654
- [ ] https://github.com/OCA/storage/pull/655
- [ ] https://github.com/OCA/storage/pull/656

`OCA/web-api` PR's that can be updated with `_preserve_not_env_managed_data` post init hook:

- [ ] https://github.com/OCA/web-api/pull/155
- [ ] https://github.com/OCA/web-api/pull/153
- [ ] https://github.com/OCA/web-api/pull/146
Add this hook as an improvement to already merged https://github.com/OCA/web-api/tree/19.0/webservice_server_env

@lmignon @sbidoul @etobella @rousseldenis @ivantodorovich WDYT?

PR #266, opened 2026-05-20
[18.0] [FIX] server_environment: no interpolation in the configparser

### LINKED ISSUE #265: [18.0] server_environment: crash when a value contains a %
## Module
server_environment

## Describe the bug
when a % is present in a value in my odoo.conf, i get this error when starting the server 
configparser.InterpolationSyntaxError: '%' must be followed by '%' or '(', found: '%rod'

## To Reproduce
**Affected versions**: at least 18
Edit server_environment/tests/testfiles/testing/base.conf, adding a % in any value. it will crash upon running the tests. 

**Expected behavior**
I want the value to be taken as is

**Additional context**
I know that the problem can be avoided by doubling the %, but interpolation should be by default disabled.

AxelPrel:
Fixes #265

### DIFF
--- a/server_environment/readme/CONFIGURE.md
+++ b/server_environment/readme/CONFIGURE.md
@@ -106,3 +106,22 @@ Note: empty environment keys always take precedence over default fields
 Read the documentation of the class
 [models/server_env_mixin.py](models/server_env_mixin.py) and [models/server_env_tech_name_mixin.py]
 (models/server_env_tech_name_mixin.py)
+
+## ConfigParser interpolation
+
+By default, ConfigParser interpolation is enabled to preserve compatibility
+
+It can be disabled:
+
+::
+
+   [options]
+   server_environment_config_interpolation = False
+
+Or with the environment variable:
+
+::
+
+   export SERVER_ENV_CONFIG_INTERPOLATION=False
+
+When enabled, standard Python ``configparser`` interpolation rules apply.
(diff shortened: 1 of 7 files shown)


sbidoul (dismissed):

AxelPrel:
yes this is a breaking change. any user of the module that explicitely use the configparser with interpolation (ie %(my_variable)) will get no interpolation at all. 
If the use of the parser is to be kept, then it must be documented in the module that this feature exists. 
And I have to find another way to bypass this problem (maybe by a configuration key in the conf that tells the module what config to give the configparser, with or without interpolation) 
If the use of the parser is just something that was forgotten when the module was made, then the change I proposed is to be kept. Your call

AxelPrel:
oh, and any user that doubled their %, bypassing this problem, will also suffer from this change, since the escape will now be taken literally, they will see two %% instead of just one % before, i think this is what you meant

the escaping solution was not approved where we use this module, since we are talking about secrets that will eventually be placed directly in the conf by an external tool (apikeys, passwords, etc) and escaping the % at this stage, even though it fixed the problem, was considered a dirty workaround that was not approved. this is why i proposed this change

sbidoul:
> oh, and any user that doubled their %, bypassing this problem, will also suffer from this change, since the escape will now be taken literally, they will see two %% instead of just one % before, i think this is what you meant

Yes exactly that, I've been hit by that this week too...

But breaking that behaviour will make things awful for those who already escaped.

So let's make this configurable with a system parameter and some documentation?

AxelPrel:
@sbidoul i have made the requested changes

houzefa-abba:
Just rebased on latest 18.0

PR #278, opened 2026-07-22
[19.0][FIX] server_environment: add search support for env-computed fields

AungKoKoLin1997:
## Summary

When `server_environment` transforms a field to an env-computed field, it sets `store=False` but does not provide a `search` method. This makes the field unusable in domain filters, e.g.:

```python
storage_id = fields.Many2one(
    comodel_name="storage.backend",
    domain=[("backend_type", "=", "sftp")],
)
```

raises:

```
ValueError: Cannot convert storage.backend.backend_type to SQL because it is not stored
```

This was not an issue before Odoo 19 because the old domain processing (`osv.expression`) handled non-stored fields more leniently. Odoo 19's rewritten domain engine (`odoo.orm.domains`) strictly requires either `store=True` or an explicit `search` method.

## Fix

Add a `search` method for env-computed fields following the same `_partialmethod` pattern already used for the `inverse` method.

@qrtl QT7028

### DIFF
--- a/server_environment/models/server_env_mixin.py
+++ b/server_environment/models/server_env_mixin.py
@@ -277,6 +277,14 @@ class ServerEnvMixin(models.AbstractModel):
                 else:
                     record._compute_server_env_from_default(field_name, options)
 
+    def _search_server_env(self, field_name, operator, value):
+        # env-computed fields are not stored, so we can't search them in SQL.
+        # Load all records and filter them in memory instead. These config
+        # models hold very few records, so the full scan is acceptable.
+        all_records = self.search([])  # pylint: disable=no-search-all
+        matched = all_records.filtered_domain([(field_name, operator, value)])
+        return [("id", "in", matched.ids)]
+
     def _inverse_server_env(self, field_name):
         options = self._server_env_fields[field_name]
         default_field = self._server_env_default_fieldname(field_name)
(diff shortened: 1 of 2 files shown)


server_environment/models/server_env_mixin.py
```diff
@@ -277,6 +277,14 @@ def _compute_server_env(self):
                 else:
                     record._compute_server_env_from_default(field_name, options)
 
+    def _search_server_env(self, field_name, operator, value):
+        # env-computed fields are not stored, so we can't search them in SQL.
+        # Load all records and filter them in memory instead. These config
+        # models hold very few records, so the full scan is acceptable.
+        all_records = self.search([])  # pylint: disable=no-search-all
```
kanda999, line 284:
```suggestion
        all_records = self.with_context(active_test=False).search([])  # pylint: disable=no-search-all
```

AungKoKoLin1997: I don't think we need to search archived records as well. Do you have any reason?

kanda999: I think it would be better to include active=False support, since this is intended to be a generic search method.

AungKoKoLin1997: Odoo standard generic search only find the active records, no?

PR #261, opened 2026-03-25
[IMP] server_environment: module uninstallation

gurneyalex:
Add a helper to manage the restoring of the database columns when a module using `server_environment` is uninstalled or the dependency on `server_environment` is dropped.

Document how to use the helper in an uninstall script or in an upgrade script (if a new version of the addon drops the dependency).

PR #275, opened 2026-06-29
[IMP] server_environment: module uninstallation

gurneyalex:
Add a helper to manage the restoring of the database columns when a module using `server_environment` is uninstalled or the dependency on `server_environment` is dropped.

Document how to use the helper in an uninstall script or in an upgrade script (if a new version of the addon drops the dependency).

This is a backward port of #261

PR #279, opened 2026-08-18
[17.0][BKP] server_environment: module uninstallation

yankinmax:
Add a helper to manage the restoring of the database columns when a module using `server_environment` is uninstalled or the dependency on `server_environment` is dropped.

Document how to use the helper in an uninstall script or in an upgrade script (if a new version of the addon drops the dependency).

Backport of:
- [ ] https://github.com/OCA/server-env/pull/261


server_environment/tests/test_environment_variable.py
```diff
@@ -1,4 +1,4 @@
-# Copyright 2018 Camptocamp (https://www.camptocamp.com).
+# Copyright 2018 2026 Camptocamp (https://www.camptocamp.com).
```
simahawk: this change is useless but if it's coming from the original work keep it 🤷

simahawk (approved):

server_environment/tests/test_environment_variable.py
yankinmax: This is just a cherry pick result

yankinmax: Actually I was thinking about restoring the original content and include only needed changes.

ivantodorovich (approved):

yankinmax:
I'm going to close this one, because have found @gurneyalex has already opened such PR (I didn't see at first). He is an original author!
- [ ] https://github.com/OCA/server-env/pull/275

PR #286, opened 2026-08-24
[17.0][BKP] server_environment: Preserve not env managed data

yankinmax:
Backport of:
- [ ] https://github.com/OCA/server-env/pull/264

### DIFF
--- a/mail_environment/migrations/17.0.1.0.2/post-migration.py
+++ b/mail_environment/migrations/17.0.1.0.2/post-migration.py
@@ -1,0 +1,9 @@
+# Copyright 2026 ACSONE SA/NV
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from openupgradelib import openupgrade
+
+
+@openupgrade.migrate()
+def migrate(env, version):
+    env["ir.mail_server"]._preserve_not_env_managed_data(["smtp_authentication"])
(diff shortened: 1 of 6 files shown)


mail_environment/migrations/17.0.1.0.2/post-migration.py
```diff
@@ -0,0 +1,9 @@
+# Copyright 2026 ACSONE SA/NV
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from openupgradelib import openupgrade
+
+
+@openupgrade.migrate()
+def migrate(env, version):
+    env["ir.mail_server"]._preserve_not_env_managed_data(["smtp_authentication"])
```
yankinmax, line 9: Set `17.0.1.0.2` as next patch one release ahead of current

simahawk (approved):

server_environment/models/server_env_mixin.py
```diff
@@ -425,3 +425,31 @@ def _setup_base(self):
             self._server_env_transform_field_to_read_from_env(field)
             self._server_env_add_is_editable_field(field)
         return
+
+    @api.model
+    def _preserve_not_env_managed_data(self, field_name_list):
```
simahawk, line 430: you should bump this module's version too as we want to merge w/ nobump

PR #288, opened 2026-08-24
[19.0][FWP] server_environment: Preserve not env managed data

yankinmax:
Forward-port of:
- [ ] https://github.com/OCA/server-env/pull/264
- [ ] https://github.com/OCA/server-env/pull/295

### DIFF
--- a/mail_environment/__manifest__.py
+++ b/mail_environment/__manifest__.py
@@ -3,7 +3,7 @@
 
 {
     "name": "Mail configuration with server_environment",
-    "version": "19.0.1.1.0",
+    "version": "19.0.1.1.1",
     "category": "Tools",
     "summary": "Configure mail servers with server_environment_files",
     "author": "Camptocamp, Odoo Community Association (OCA)",
--- a/mail_environment/migrations/19.0.1.1.1/post-migration.py
+++ b/mail_environment/migrations/19.0.1.1.1/post-migration.py
@@ -1,0 +1,9 @@
+# Copyright 2026 ACSONE SA/NV
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from openupgradelib import openupgrade
+
+
+@openupgrade.migrate()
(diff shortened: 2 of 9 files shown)


mail_environment/migrations/19.0.1.1.1/post-migration.py
```diff
@@ -0,0 +1,9 @@
+# Copyright 2026 ACSONE SA/NV
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from openupgradelib import openupgrade
+
+
+@openupgrade.migrate()
+def migrate(env, version):
+    env["ir.mail_server"]._preserve_not_env_managed_data(["smtp_authentication"])
```
yankinmax, line 9: Set `19.0.1.1.1` as next patch one release ahead of current.

mail_environment/__manifest__.py
```diff
@@ -3,7 +3,7 @@
 
 {
     "name": "Mail configuration with server_environment",
-    "version": "19.0.1.1.0",
+    "version": "19.0.1.1.1",
```
yankinmax, line 6: `pre-commit` argues if manifest version is not changed

simahawk: just to be clear: is kind of mandatory to change the version if you provide an upgrade script specific to a version. The problem arises only if you change 2 modules and you bump only 1 module version.
In such cases we should bump both to be able to merge w/ `nobump`.
We might need to update the guidelines :thinking:

simahawk (approved):

mail_environment/__manifest__.py
yankinmax: I've drooped this comment before we've discussed I need to mainly update the version.

ivantodorovich (approved):

yankinmax:
Hello @simahawk your commits are cherry-picked from the:
- [ ] https://github.com/OCA/server-env/pull/295

PR #292, opened 2026-08-31
[FIX] server_environment: fix deadlock between gevent and subprocesses

gurneyalex:
Call to `subprocess.Popen` caused a deadlock. This broke the gevent worker when `workers > 0`. As a consequence, the port 8072 supposed to handle /websocket never responds.

Gevent warns against it on startup:
> gevent/os.py:426: DeprecationWarning: This process is multi-threaded, use of fork() may lead to deadlocks in the child.

This is a backport of #272


simahawk:
@gurneyalex mind to have a look at tests?

PR #295, opened 2026-09-04
[18.0][FIX] server_environment: _preserve_not_env_managed_data

simahawk:
Preserve default values already stored in the DB.


ivantodorovich (approved):

lmignon (approved):

lmignon (commented):
a unittest would have been welcome 😏

simahawk:
> a unittest would have been welcome 😏

Added!

yankinmax (approved):

yankinmax:
Thank you @simahawk 
I'll plan to BKP and FWP this PR and then we can proceed to test on storage and webservice opened PR's and finally get them merged.
But, as it's an update for the server-env possibly we don't need to update anything.

